PR #2, opened 2024-05-01
Take default values into account

DmitrySharabin:
For now, when we apply changes to a property on its change, we only consider authored properties. If I get the logic correctly, when `change.source` equals `default`, we should use the default value (from the prop spec) if it is provided.

@LeaVerou, could you please take a look at the proposed change? Do I get the logic correctly, or does this fix the `<channel-slider>` component by accident? 😅

### DIFF
--- a/src/common/Prop.js
+++ b/src/common/Prop.js
@@ -186,7 +186,10 @@ export default class Prop {
 			}
 		}
 		else if (change.source === "property" || change.source === "default") {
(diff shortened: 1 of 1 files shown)


LeaVerou (approved):
LGTM, does it fix the issue?

DmitrySharabin:
> LGTM, does it fix the issue?

It does.

PR #7, opened 2024-05-02
First pass on fixing <color-picker>

DmitrySharabin:
- When the component is mounted, we must dispatch the `input` event to update the corresponding color swatch.
- Previously, for the `color` prop, we used the corresponding spec from the `<channel-slider>`. But it depends on the `channel` property `<color-picker>` doesn't have. I'm sure the solution I suggest is not optimal at all. I just needed something to make the picker work.


DmitrySharabin:
By default, the color swatch provided by `<color-picker>` is editable. Should the picker listen to the swatch's `colorchange` event and adjust the sliders accordingly (and avoid the loop caused by those changes)?

LeaVerou (approved):

PR #9, opened 2024-05-03
Add ESLint

DmitrySharabin:
Based on [`.eslintrc.cjs`](https://github.com/color-js/color.js/blob/main/.eslintrc.cjs) from the `color.js` repo. Uses ESLint `v9`.


LeaVerou:
I would have been fine with a direct commit for this btw

PR #10, opened 2024-05-03
First pass at colorchange event

LeaVerou:
Why on earth is this not working?

### DIFF
--- a/src/color-slider/README.md
+++ b/src/color-slider/README.md
@@ -24,7 +24,7 @@ You can listen to the `input` event and grab the `color` property to get the cur
 ```html
 <color-slider space="hsl"
               stops="red, yellow, lime, aqua, blue, magenta, red"
-              oncolorchange="console.log(this.nextElementSibling.textContent = this.color)"></color-slider>
+              oncolorchange="this.nextElementSibling.textContent = this.color"></color-slider>
 <output></output>
 ```
 
--- a/src/color-slider/README.md
+++ b/src/color-slider/README.md
@@ -19,7 +19,7 @@ Basic example:
               stops="oklch(90% 50% 100), darkcyan, indigo"></color-slider>
 ```
 
-You can listen to the `input` event and grab the `color` property to get the current color value:
+You can listen to the `colorchange` event and grab the `color` property to get the current color value:
 
 ```html
 <color-slider space="hsl"
--- a/src/color-slider/README.md
+++ b/src/color-slider/README.md
@@ -24,7 +24,7 @@ You can listen to the `input` event and grab the `color` property to get the cur
 ```html
 <color-slider space="hsl"
               stops="red, yellow, lime, aqua, blue, magenta, red"
-              oninput="this.nextElementSibling.textContent = this.color"></color-slider>
+              oncolorchange="console.log(this.nextElementSibling.textContent = this.color)"></color-slider>
 <output></output>
 ```
 
@@ -33,7 +33,7 @@ In fact, you can combine it with a [`<color-inline>`](../color-inline/) or [`<co
 ```html
 <color-slider space="oklch"
               stops="oklch(80% 50% 70), oklch(65% 50% 180)"
-              oninput="this.nextElementSibling.textContent = this.color"></color-slider>
+              oncolorchange="this.nextElementSibling.textContent = this.color"></color-slider>
 <color-inline></color-inline>
 ```
 
@@ -43,7 +43,7 @@ You can set the `value` attribute to specify an initial color other than the mid
 <color-slider space="oklch"
               stops="oklch(85% 50% 80), oklch(65% 50% 180)"
 			  value="0.1"
-              oninput="this.nextElementSibling.textContent = this.color"></color-slider>
+              oncolorchange="this.nextElementSibling.textContent = this.color"></color-slider>
 <color-inline></color-inline>
 ```
 
@@ -53,7 +53,7 @@ You can use a different min and max value and it’s just linearly mapped to the
 <color-slider space="oklch"
               stops="oklch(85% 50% 80), oklch(65% 50% 180)"
 			  min="-50" max="50" value="20"
-              oninput="this.nextElementSibling.textContent = this.color"></color-slider>
+              oncolorchange="this.nextElementSibling.textContent = this.color"></color-slider>
 <color-inline></color-inline>
 ```
 
@@ -64,7 +64,7 @@ You can add an editable tooltip by simply using the `tooltip` attribute:
               stops="oklch(85% 50% 80), oklch(65% 50% 180)"
 			  min="-50" max="50" value="20"
               tooltip
-              oninput="this.nextElementSibling.textContent = this.color"></color-slider>
+              oncolorchange="this.nextElementSibling.textContent = this.color"></color-slider>
 <color-inline></color-inline>
 ```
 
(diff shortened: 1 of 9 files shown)


src/color-slider/README.md
```diff
@@ -126,7 +136,8 @@ If you’re only using the CSS file, you should set these yourself.
 
 | Name | Description |
 |------|-------------|
-| `input` | Fired when the color changes and once during initialization |
+| `input` | Fired when the color changes due to user action |
+| `change` | Fired when the color changes for any reason, and once during initialization |
```
DmitrySharabin, line 140: Do you mean `colorchange`?

src/color-slider/color-slider.js
```diff
@@ -37,7 +38,7 @@ export default class ColorSlider extends HTMLElement {
 
 			this.#initialized = true;
 
-			this._el.slider.dispatchEvent(new Event("input"));
+			this._el.slider.dispatchEvent(new Event("colorchange"));
```
DmitrySharabin, line 41:
```suggestion
			this._el.slider.dispatchEvent(new ColorChangeEvent("colorchange", {bubbles: true, color: this.color}));
```

DmitrySharabin: Based on my experiments, this (partially) fixes the issue.

DmitrySharabin:
I also noticed that <color-sliders> with “standard” attributes (`min`, `max`, `value`) are broken (they don’t handle the `colorchange` event). If you remove them, they start working as they should.

src/color-slider/color-slider.js
```diff
@@ -260,7 +287,7 @@ defineFormAssociated(ColorSlider, {
 	getSource: el => el._el.slider,
 	role: "slider",
 	valueProp: "value",
-	changeEvent: "input",
+	changeEvent: "change",
```
DmitrySharabin, line 290: Is this an additional event to the ones we already have (`input` and `colorchange`)?

src/color-slider/color-slider.js
LeaVerou: Gah, I knew it must be something silly like that. Thanks Dmitry!

src/color-slider/color-slider.js
LeaVerou: Neither would be correct actually. This is the event that updates this control's form associated value, which is `this.value`. I think I need to convert `Prop` to using an event instead of the callback (which I've been wanting to do anyway), then `defineformAssociated()` can just listen to prop changes.

src/color-slider/README.md
LeaVerou: At the time I wrote this I didn't didn't, but now yes. Instead, we should also support `change`, but fire it when the slider does. I'm wondering if we should have a `valuechange` event as well.

Alternatively, we could get rid of both of them and just use `propchange` 🤔

src/color-slider/color-slider.js
DmitrySharabin: Please correct me if I'm wrong. Having this, let me use `<color-slider>` in my Mavo app, and all I need to do to make it work is to teach Mavo that this element should be handled as a form control, like so

```html
<color-slider property="color" mv-edit-as="formNumber" space="oklch" stops="gold, darkcyan, indigo"></color-slider>
```

And this should work, right?

src/color-slider/README.md
DmitrySharabin: > Alternatively, we could get rid of both of them and just use `propchange` 🤔

IMHO, this can help if one day we decide that users should also be notified (and able to listen to) about all sorts of events, e.g., `spacechange`, `stopschange`, etc.

src/color-slider/color-slider.js
```diff
@@ -30,15 +32,17 @@ export default class ColorSlider extends HTMLElement {
 		this._el.slider.addEventListener("input", this);
 		this._el.spinner.addEventListener("input", this);
```
DmitrySharabin, line 33: I wonder if we should move these lines to `connectedCallback()` since we remove the corresponding listeners inside `disconnectedCallback()`? 🤔

src/color-slider/color-slider.js
```diff
@@ -114,6 +117,15 @@ export default class ColorSlider extends HTMLElement {
...
 		}
+
+		if (name === "color" || name === "defaultColor") {
+			let color = this.color;
+
+			if (color) {
+				let displayedColor = color.display();
+				this.style.setProperty("--color", displayedColor);
+			}
+		}
```
DmitrySharabin, line 128: This is a duplicate of lines 103–110 above. Do we need this logic twice, or should we remove either (I don't see why the order matters)?

DmitrySharabin (changes requested):
See comments. There are few.

Awesome changes!

src/color-slider/README.md
```diff
@@ -24,7 +24,7 @@ You can listen to the `input` event and grab the `color` property to get the cur
 ```html
 <color-slider space="hsl"
               stops="red, yellow, lime, aqua, blue, magenta, red"
-              oninput="this.nextElementSibling.textContent = this.color"></color-slider>
+              oncolorchange="this.nextElementSibling.textContent = this.color"></color-slider>
```
DmitrySharabin, line 27: I'm writing it here since line 22 is not in the PR.

Do I get it correctly that we should mention the `colorchange` event instead of `input` in line 22?

src/common/Props.js
```diff
@@ -97,21 +121,66 @@ export default class Props extends Map {
...
 
-		// Build or update dependency graph
+		// Rebuild dependency graph
+		let dependents = {};
+
+		for (let name of this.keys()) {
+			dependents[name] = new Set();
+		}
+
+		let keyIndices = Object.fromEntries(this.keys().map((key, i) => [key, i]));
```
DmitrySharabin, line 153: For some reason, this line throws an error in Firefox: `Uncaught TypeError: this.keys().map is not a function`.

Safari fails silently. 🤬

src/color-slider/color-slider.js
LeaVerou: Sounds good

src/color-slider/README.md
LeaVerou: Both should be mentioned

src/common/Props.js
DmitrySharabin:
```suggestion
		let keyIndices = Object.fromEntries([...this.keys()].map((key, i) => [key, i]));
```

DmitrySharabin: It does! Yay! 🥳

Suggested the change.

LeaVerou: Yeah, that's because adding `map` and friends to iterables is pretty new

src/color-slider/README.md
LeaVerou: Oh I see what you mean. Yes, the prose should change

src/color-slider/color-slider.js
DmitrySharabin: If you like, I can commit and push this and other changes we agree on.

src/color-slider/color-slider.js
LeaVerou: Can't remember what `formNumber` is, you'd need to pick the one that `<input type=range>` is under.

src/color-slider/color-slider.js
LeaVerou: Yes please!

src/color-slider/color-slider.js
DmitrySharabin: Awesome! Thanks.

src/color-slider/README.md
DmitrySharabin: How about this? Should we fix the `Events` table by changing `change` to `colorchange` and adding the `valuechange` event?

LeaVerou: Yes. We should also have a change event. Can you check if the change event from the slider propagates by listening to it from the outside? I've been meaning to check this for a while. 
I wonder if we should also list `propchange`. In which case, we could add an `onpropchange` prop too.

LeaVerou: Btw no need to send PRs for commits that just *add* documentation

PR #11, opened 2024-05-06
propChangedCallback → propchange event

LeaVerou:
First pass at converting propChangedCallback to a more composable propchange event

### DIFF
--- a/src/channel-slider/channel-slider.js
+++ b/src/channel-slider/channel-slider.js
@@ -24,6 +24,7 @@ export default class ChannelSlider extends HTMLElement {
 
 		this._el = dom.named(this);
 		this._el.slot = this.shadowRoot.querySelector("slot");
+		this.addEventListener("propchange", this.propChangedCallback);
 	}
 
 	connectedCallback() {
@@ -88,9 +89,7 @@ export default class ChannelSlider extends HTMLElement {
 		return this._el.slider.progressAt(p);
 	}
 
-	propChangedCallback (prop, change) {
-		let name = prop.name;
-
+	propChangedCallback ({name, prop, detail: change}) {
 		if (["space", "min", "max", "step", "value", "defaultValue"].includes(name)) {
 			prop.applyChange(this._el.slider, change);
 		};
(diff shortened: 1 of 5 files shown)


src/color-picker/color-picker.js
```diff
@@ -26,6 +26,7 @@ const Self = class ColorPicker extends HTMLElement {
 		`;
 
 		this._el = dom.named(this);
+		this.addEventListener("propchange", this.propChangedCallback);
```
DmitrySharabin, line 29: Shouldn't we remove the corresponding listener in `disconnectedCallback()` here and in other components? Or will this listener be destroyed automatically when the component is disconnected?

Can't add this as a suggestion since `disconnectedCallback()`'s lines are not in the PR.

src/channel-slider/channel-slider.js
```diff
@@ -88,9 +89,7 @@ export default class ChannelSlider extends HTMLElement {
 		return this._el.slider.progressAt(p);
 	}
 
-	propChangedCallback (prop, change) {
-		let name = prop.name;
-
+	propChangedCallback ({name, prop, detail: change}) {
```
DmitrySharabin, line 92: We rename `detail` to `change` here, right?

DmitrySharabin (commented):
I have a couple of questions. :)

src/color-picker/color-picker.js
LeaVerou: I don't think so since we're adding this in the constructor, not in `connectedCallback()`. Do we only want it to run when the component is connected?

DmitrySharabin: I meant, where should we remove the corresponding event listener in that case? We don't do it anywhere now.

LeaVerou: Why do we need to remove it?

DmitrySharabin: I thought it's a good practice to remove event listeners when they are no longer needed. If only they were removed automatically somehow. I can see that the listener is added inside the constructor. Does it mean we don't need to remove such listeners? Honestly, I don't know the answer to this question. 🤷‍♂️

LeaVerou: When is it no longer needed? What is the point when we no longer need to listen to prop changes?

DmitrySharabin: I guess we don't need to remove the listener. :)

DmitrySharabin (approved):

PR #21, opened 2024-05-07
First stab at base class + events helper

LeaVerou:


src/common/Element.js
```diff
@@ -0,0 +1,53 @@
...
+			this.initializeProps?.();
+
+			this.#initialized = true;
+		}
+	}
+
+
+	static init () {
+		// One time stuff
+		if (this._initialized) {
```
DmitrySharabin, line 33: Is it some sort of static variable that shouldn't be defined? I don't understand how we can use uninitialized variables in static methods. I didn't find it `_initialized` in other parts of the code base. Does the `HTMLElement` class (instance) have it?

src/common/PropChangeEvent.js
```diff
@@ -7,52 +7,41 @@ export class PropChangeEvent extends CustomEvent {
...
+		Object.defineProperty(CustomPropChangeEvent, "name", { value: eventClassName });
+		return CustomPropChangeEvent;
+	}
+}
 
-			this[name] = options[name];
-		}
-	};
-	Object.defineProperty(CustomPropChangeEvent, "name", { value: eventClassName });
+export default function definePropChangeEvent (Class, name) {
```
DmitrySharabin, line 36: Looking at the implementation, I can't see why we need the `Class` parameter—it's not used anywhere. Does this helper can be used without a class at all?

LeaVerou: Good point, removed!

src/common/Element.js
```diff
@@ -0,0 +1,53 @@
...
+
+	}
+
+	connectedCallback () {
+		if (!this.#initialized) {
+			this.initializeProps?.();
+
+			this.#initialized = true;
+		}
+	}
```
DmitrySharabin, line 28: For some reason, this code is not reachable. Because of this, `<color-slider>` is broken.

DmitrySharabin: Should we call `super.connectedCallback()` from the subclasses?

DmitrySharabin: Is it legal at all? :)

LeaVerou: > Should we call `super.connectedCallback()` from the subclasses?

Oh yes, absolutely. I must have forgotten

src/common/Element.js
DmitrySharabin: Got it. Thanks!

PR #23, opened 2024-05-08
Remove no longer necessary code

DmitrySharabin:
We no longer need to fire the `input` event on initialization—everything is handled automatically by `valuechange`/`colorchange` event listeners.

This makes the `#initialized` property redundant and fixes the weirdness when the sider has two private `#initialized` properties (one inherited from the `NudeElement` class).


LeaVerou (approved):
Sweet, it was on my to-do list to check if that's the case, thanks!

PR #24, opened 2024-05-08
Fix typo in `defineEvent()`

DmitrySharabin:
Fixes the check whether `oneventname` is implemented in a class.


LeaVerou (approved):
🤦🏽‍♀️

thanks!

PR #28, opened 2024-05-08
Fix post construct stuff

DmitrySharabin:


LeaVerou (approved):
Glad this fixed it. What's very strange is that I'd swear I tried this at some point yesterday with no luck. Anyway, LGTM!

Issue #16, opened 2024-05-06
`tooltip="progress"` on `<color-slider>` (and other related elements) inherits `min`, `max`, and `step` from the slider


DmitrySharabin:
Instead of setting them to `0`, `100`, and `1`, respectively.

We can move the slider thumb and see the correct result in the tooltip, but changing the progress manually might be impossible.

Testcase

```html
<color-slider
	space="oklch"
	stops="oklch(85% 50% 80), oklch(65% 50% 180)"
	min="-50" max="50" value="20"
	tooltip="progress"
></color-slider>
```

PR #33, opened 2024-05-15
First attempt to fix issue #16

DmitrySharabin:
For now, it doesn't fix anything—what was broken previously is still broken.
Why? At the element initialization, `this.tooltip` is `undefined`, so the spinner will get its values for `min`, `max`, and `step` from the corresponding properties of the slider.

Currently, after `<color-slider>` is initialized, `min`, `max`, and `step` remain unchanged, which means the corresponding values of the spinner aren't being updated. I'm trying to find a solution to this.

@LeaVerou Is there a proper way to call `propChangedCallback()`(or something else) for `min`, `max`, and `step` after the `<color-slider>` is initialized. I suspect it's possible to do from `connectedCallback()`, but I'm unsure how to do it properly.

### DIFF
--- a/src/color-slider/color-slider.js
+++ b/src/color-slider/color-slider.js
@@ -63,7 +63,15 @@ const Self = class ColorSlider extends NudeElement {
 		if (["min", "max", "step", "value", "defaultValue"].includes(name)) {
 			prop.applyChange(this._el.slider, change);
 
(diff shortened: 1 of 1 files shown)


DmitrySharabin:
I'm closing this in favor of a better (from my perspective) fix: #88.

Issue #36, opened 2024-05-15
Harmonize APIs and implementations of different components


LeaVerou:
Currently, there are many differences between the older `<color-swatch>`, `<color-inline>`, `<color-gamut>` components and the newer `<color-picker>`, `<channel-slider>`, `<color-slider>` ones.

We should harmonize them.
This means:

- [ ] Using props for all instead of ad hoc implementations (unless really not needed)
- [ ] `colorchange` and `valuechange` events for all
- What else?

PR #37, opened 2024-05-15
First stab at oncolorchange for `<color-swatch>`

LeaVerou:
Related to #36 

Not working. Need to debug why.

### DIFF
--- a/src/color-swatch/color-swatch.js
+++ b/src/color-swatch/color-swatch.js
@@ -5,6 +5,8 @@ import "../color-gamut/color-gamut.js";
 let importIncrementable;
 
 const Self = class ColorSwatch extends HTMLElement {
+	static initQueue = [];
+
 	#dom = {};
 
 	constructor () {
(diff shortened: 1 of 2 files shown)


src/color-swatch/color-swatch.js
```diff
@@ -252,9 +253,16 @@ const Self = class ColorSwatch extends HTMLElement {
 		}
 	}
 
+	static events = {
+		colorchange: {},
+		valuechange: {},
```
DmitrySharabin, line 258:
```suggestion
		colorchange: {
			propchange: "color",
		},
		valuechange: {
			propchange: "value",
		},
```

DmitrySharabin: We should tell explicitly what props to monitor.

LeaVerou:
Thanks! Does it work with this change?

DmitrySharabin:
> Thanks! Does it work with this change?

It does. 🥳

PR #44, opened 2024-05-17
Some fixes

DmitrySharabin:
One thing is left: after updating `space`, `progress` is not updated. Investigating...

That's why I submit it as a draft.

### DIFF
--- a/src/channel-slider/channel-slider.js
+++ b/src/channel-slider/channel-slider.js
@@ -179,6 +179,13 @@ export default class ChannelSlider extends NudeElement {
 			type: Number,
 			default () {
 				return this.defaultColor.get(this.channel);
+				try {
+					return this.defaultColor.get(this.channel);
+				}
+				catch {
+					let channel = Object.keys(this.space.coords)[0];
+					return this.defaultColor.get(channel);
+				}
 			},
 			reflect: {
 				from: "value",
--- a/src/channel-slider/channel-slider.js
+++ b/src/channel-slider/channel-slider.js
@@ -182,8 +182,7 @@ export default class ChannelSlider extends NudeElement {
 					return this.defaultColor.get(this.channel);
 				}
 				catch {
-					let channel = Object.keys(this.space.coords)[0];
-					return this.defaultColor.get(channel);
+					return this.defaultColor.coords[0];
 				}
 			},
 			reflect: {
--- a/src/channel-slider/channel-slider.js
+++ b/src/channel-slider/channel-slider.js
@@ -178,7 +178,6 @@ export default class ChannelSlider extends NudeElement {
 		defaultValue: {
 			type: Number,
 			default () {
-				return this.defaultColor.get(this.channel);
 				try {
 					return this.defaultColor.get(this.channel);
 				}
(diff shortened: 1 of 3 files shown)


src/channel-slider/channel-slider.js
```diff
@@ -178,7 +178,13 @@ export default class ChannelSlider extends NudeElement {
 		defaultValue: {
 			type: Number,
 			default () {
-				return this.defaultColor.get(this.channel);
+				try {
+					return this.defaultColor.get(this.channel);
+				}
+				catch {
+					let channel = Object.keys(this.space.coords)[0];
```
LeaVerou, line 185: A comment here about what it means to be in this branch?

src/color-picker/color-picker.js
```diff
@@ -50,7 +50,7 @@ const Self = class ColorPicker extends NudeElement {
 			// From sliders
 			let coords = [...this._el.sliders.children].map(el => el.value);
 			this.color = new Color(this.space, coords);
-			this._el.swatch.value = this.color;
+			this._el.swatch.color = this.color;
```
LeaVerou, line 53: Can you just commit this directly since it's an obvious fix and we don't want the PR to hold it up?

src/color-picker/color-picker.js
```diff
@@ -80,6 +80,10 @@ const Self = class ColorPicker extends NudeElement {
 					this._el.sliders.insertAdjacentHTML("beforeend", `<channel-slider space="${ this.space.id }" channel="${ channel }"></channel-slider>`);
 				}
 			}
+
+			if (change.oldInternalValue || change.oldAttributeValue) {
+				this.render();
```
LeaVerou, line 85: Why? A comment would be helpful here!

src/color-slider/color-slider.js
```diff
@@ -99,10 +99,10 @@ const Self = class ColorSlider extends NudeElement {
 				this.style.setProperty("--color", displayedColor);
 			}
 		}
-		else if (name === "value") {
+		else if (name === "value" || name === "min" || name === "max") {
```
LeaVerou, line 102: This could also be a direct commit.

src/channel-slider/channel-slider.js
```diff
@@ -178,7 +178,13 @@ export default class ChannelSlider extends NudeElement {
...
+					return this.defaultColor.get(channel);
```
LeaVerou, line 186: If we're just trying to get the first coord, it’s easier to do `this.defaultColor.coords[0]`.

src/color-slider/color-slider.js
DmitrySharabin: Let me commit this change and the one above (with a fix for the typo) directly and rebase this branch so we can iterate.

src/channel-slider/channel-slider.js
DmitrySharabin: Oh, yes. Much easier.  Thanks!

src/color-picker/color-picker.js
DmitrySharabin: Sure!

src/color-picker/color-picker.js
DmitrySharabin: When we first come here on component initialization, there is no need to call `render()` since it'll be rendered automatically after it's connected. So, we only need to re-render the component when `space` actually changes _after_ the component is mounted.

There were no issues when I called `render()` without any condition, but I thought I could optimize it a bit by not re-rendering the component when there is no need.

DmitrySharabin:
Close this issue since I broke it into two separate PRs:
- https://github.com/color-js/elements/pull/47
- https://github.com/color-js/elements/pull/46

PR #46, opened 2024-05-20
[channel-slider] Don't throw on unknown channels when working with `defaultValue`

DmitrySharabin:
We face this issue when we try to set `space` programmatically (e.g., in `<color-picker>`).

### DIFF
--- a/src/channel-slider/channel-slider.js
+++ b/src/channel-slider/channel-slider.js
@@ -178,7 +178,15 @@ export default class ChannelSlider extends NudeElement {
 		defaultValue: {
 			type: Number,
 			default () {
-				return this.defaultColor.get(this.channel);
+				try {
+					return this.defaultColor.get(this.channel);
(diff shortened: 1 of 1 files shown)


DmitrySharabin:
`<color-picker>` works correctly (after merging #47) even if this PR is not merged. The only thing that might be a bit annoying is that we have errors in the console when setting `space` on `<color-picker>` programmatically.

image

LeaVerou (approved):
Approving since this is an improvement over the current state but do note that this could become a problem if down the line other props depend on `defaultValue`.
Also, does this produce better results than simply returning `0`? (besides being right about 1/3rd of the time) If we just need a value that doesn't have dependencies, `0` seems to fit the bill.

DmitrySharabin:
Nah. Let's not merge it if it's not good enough. We'll find a better solution. :)

PR #49, opened 2024-05-20
[color-slider] Fix active style in WebKit-based browsers

DmitrySharabin:
For some reason, the `&::-webkit-slider-thumb:active` rule doesn't work.


LeaVerou (approved):

LeaVerou:
Maybe add a comment so we don't move it back in when refactoring

PR #51, opened 2024-05-21
[color-picker] Fix regression when provided color was replaced by the default one on component initialization

DmitrySharabin:
When we render `<color-picker>` in the process of its initialization (on `space` change), we replace the color (provided via the `color` attribute) with the default one. We definitely don't want it. I noticed this earlier, but yesterday, it slipped my mind. Sorry about that. This PR should fix the regression.

### DIFF
--- a/src/color-picker/color-picker.js
+++ b/src/color-picker/color-picker.js
@@ -30,6 +30,8 @@ const Self = class ColorPicker extends NudeElement {
 		super.connectedCallback?.();
 		this._el.sliders.addEventListener("input", this);
 		this._el.swatch.addEventListener("input", this);
+
+		this.render();
 	}
 
 	disconnectedCallback() {
@@ -79,7 +81,12 @@ const Self = class ColorPicker extends NudeElement {
 				}
 			}
 
-			this.render();
+			// When we first come here on component initialization,
+			// we shouldn't call render() since it'll replace the provided color with the default one.
+			// So, we only need to re-render the component when space actually changes *after* the component is mounted.
+			if (change.oldInternalValue || change.oldAttributeValue) {
+				this.render();
+			}
 		}
 
 		if (name === "color") {


LeaVerou:
This sounds a bit like a workaround rather than a fix at first glance. `render()` should not be replacing the current color with the default one, that sounds like a bug in `render()`. Could you tell me a bit more about what happens?

DmitrySharabin:
Sure.

On component initialization, `space` is set, and `render()` is called, the following block is executed:

https://github.com/color-js/elements/blob/e8ec6d801de8b31dc8bf3a9cda34a13ef1817d6d/src/color-picker/color-picker.js#L48-L51

In this block, a new color is being built based on the channel sliders' values (line 48), i.e., the default ones. Then, it is _set_ in line 49. The corresponding setter updates the `color` attribute, and we lose the provided color _before_ it was read and applied.

Hope it makes sense.

LeaVerou:
This whole render method feels like a relic; we shouldn't need it, prop defaults and propChangeCallback() should take care of these things in a better way.

DmitrySharabin:
> This whole render method feels like a relic; we shouldn't need it, prop defaults and propChangeCallback() should take care of these things in a better way.

Agreed.

Thinking about this issue a bit more, I recalled that we face this issue (and similar issues in other components) because `propChangeCallback()` might be called even though the component hasn't been fully initialized (connected) yet.

For example, when we set `min`, we should also set it on the spinner. However, the value that `min` should have on the spinner depends on the value of `tooltip`: if `this.tooltip === "progress"`, `min` should always be equal to `0` regardless. Unfortunately, `this.tooltip` might be `undefined` if the corresponding attribute (prop) hasn't been handled by the moment `min` is set.

I wonder if it is possible to queue all calls of `propChangeCallback()` (while the component is being initialized) and perform them one by one _after_ the component is connected (when we have all the information we need). Some of those calls probably cancel the previous ones, but it's not a big deal (at least for now).

LeaVerou:
That is an excellent point — can you file an issue on Nude Element?

DmitrySharabin:
I'm closing this one in favor of #87. There, I implemented ideas mentioned in this PR, e.g., getting rid of the relic `render()`.

PR #54, opened 2024-05-21
[color-slider] Fix the spinner `min`, `max`, and `step` values

DmitrySharabin:
Previously, if specified on `<color-slider>`, they all became equal to the slider value, not the corresponding prop's value.
This PR fixes it.

### DIFF
--- a/src/color-slider/color-slider.js
+++ b/src/color-slider/color-slider.js
@@ -63,7 +63,10 @@ const Self = class ColorSlider extends NudeElement {
 		if (["min", "max", "step", "value", "defaultValue"].includes(name)) {
 			prop.applyChange(this._el.slider, change);
 
-			let spinnerValue = this.tooltip === "progress" ? +(this.progress * 100).toPrecision(4) : change.value;
+			let spinnerValue = change.value;
+			if (this.tooltip === "progress" && (name === "value" || name === "defaultValue")) {
+				spinnerValue = +(this.progress * 100).toPrecision(4);
+			}
 			prop.applyChange(this._el.spinner, {...change, value: spinnerValue});
 		}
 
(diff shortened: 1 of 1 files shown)


src/color-slider/color-slider.js
```diff
@@ -63,7 +63,7 @@ const Self = class ColorSlider extends NudeElement {
 		if (["min", "max", "step", "value", "defaultValue"].includes(name)) {
 			prop.applyChange(this._el.slider, change);
 
-			let spinnerValue = this.tooltip === "progress" ? +(this.progress * 100).toPrecision(4) : this.value;
+			let spinnerValue = this.tooltip === "progress" ? +(this.progress * 100).toPrecision(4) : change.value;
```
LeaVerou, line 66: This will still fail if `this.tooltip` is `"progress"`. This entire code path is misguided, this line assumes you're propagating `value`, but you're actually propagating min, max, and step too.
I think we need to branch based on whether the property was min/max/step or value/defaultValue.

LeaVerou (changes requested):

src/color-slider/color-slider.js
DmitrySharabin: Absolutely! I should trust my instincts more—I rejected my first fix in favor of this one (and made a mistake). 🤦‍♂️

src/color-slider/color-slider.js
```diff
@@ -63,7 +63,10 @@ const Self = class ColorSlider extends NudeElement {
 		if (["min", "max", "step", "value", "defaultValue"].includes(name)) {
 			prop.applyChange(this._el.slider, change);
 
-			let spinnerValue = this.tooltip === "progress" ? +(this.progress * 100).toPrecision(4) : this.value;
+			let spinnerValue = change.value;
```
LeaVerou, line 66: Nit: I would call it just `value`, not `spinnerValue`.

LeaVerou (approved):

src/color-slider/color-slider.js
DmitrySharabin: Fixed. Thanks!

PR #56, opened 2024-05-21
[color-inline] Add reference tables and a demo to the docs

DmitrySharabin:


src/color-inline/README.md
```diff
@@ -28,4 +34,38 @@ Invalid color:
...
\ No newline at end of file
+```
+
+## Reference
+
+### Slots
+
+| Name | Description |
+|------|-------------|
+| (default) | The element's main content—the color to be shown. Placed next to the color swatch. |
```
LeaVerou, line 45: Italic for (default)

src/color-inline/README.md
```diff
@@ -14,6 +14,12 @@ Basic use:
 ```
 </html-demo>
 
+The color swatch only:
```
LeaVerou, line 17:
```suggestion
You can use `value` to set the color swatch while displaying something else as the content (or even nothing at all):
```

src/color-inline/README.md
```diff
@@ -28,4 +34,38 @@ Invalid color:
...
+| Name | Description |
+|------|-------------|
+| (default) | The element's main content—the color to be shown. Placed next to the color swatch. |
+
+### Attributes & Properties
+
+| Attribute | Property | Property type | Default value | Description |
+|-----------|----------|---------------|---------------|-------------|
+| `color` | `color` | `Color` &#124; `null` | - | The current color value. `null` for invalid colors. |
+| `value` | `value` | `string` &#124; `undefined` | - | The current value. `undefined` if the `value` attribute is not set. |
```
LeaVerou, line 52: It’s not helpful to document `value` as "the current value", that means nothing. It's as informative as the old `i++; // increment i` joke. 

Explain that it’s the textual form of the color, and will have a value even if the color is invalid.

src/color-inline/README.md
```diff
@@ -28,4 +34,38 @@ Invalid color:
...
+
+
+### CSS variables
+
+| Variable | Type | Description |
+|----------|---------------|-------------|
+| `--transparency-grid` | `<image>` | Gradient used as a background for transparent parts of the swatch. |
+| `--transparency-tile-size` | `<length>` | The size of the tiles in the transparency grid. |
```
LeaVerou, line 60:
```suggestion
| `--transparency-tile-size` | `<length>` | The size of the tiles in the transparency grid. This will not be used if you are overriding `--transparency-grid` |
```

LeaVerou: Also, can we make sure these are harmonized across all components that have a transparency grid?

src/color-inline/README.md
```diff
@@ -28,4 +34,38 @@ Invalid color:
...
+| `--border-width` | `<length>` | The width of the border around the swatch. |
+| `--box-shadow-blur` | `<length>` | The blur radius of the box shadow around the swatch. |
+| `--box-shadow-color` | `<color>` | The color of the box shadow around the swatch. |
+| `--color-image` | `<image>` | The image used to render the swatch. Will override the image that depends on the actual color, so you should rarely use this directly. |
```
LeaVerou, line 64: Yeah let’s not document that.

src/color-inline/README.md
```diff
@@ -28,4 +34,38 @@ Invalid color:
...
+
+### CSS variables
+
+| Variable | Type | Description |
+|----------|---------------|-------------|
+| `--transparency-grid` | `<image>` | Gradient used as a background for transparent parts of the swatch. |
+| `--transparency-tile-size` | `<length>` | The size of the tiles in the transparency grid. |
+| `--border-width` | `<length>` | The width of the border around the swatch. |
+| `--box-shadow-blur` | `<length>` | The blur radius of the box shadow around the swatch. |
+| `--box-shadow-color` | `<color>` | The color of the box shadow around the swatch. |
```
LeaVerou, line 63: Yikes, are these really completely unprefixed? Oh well.

LeaVerou (approved):

LeaVerou:
Accidentally approved; see comments.

src/color-inline/README.md
```diff
@@ -42,14 +42,14 @@ Invalid color:
...
-| (default) | The element's main content—the color to be shown. Placed next to the color swatch. |
+| _(default)_ | The element's main content—the color to be shown. Placed next to the color swatch. |
 
 ### Attributes & Properties
 
 | Attribute | Property | Property type | Default value | Description |
 |-----------|----------|---------------|---------------|-------------|
 | `color` | `color` | `Color` &#124; `null` | - | The current color value. `null` for invalid colors. |
-| `value` | `value` | `string` &#124; `undefined` | - | The current value. `undefined` if the `value` attribute is not set. |
+| `value` | `value` | `string` &#124; `undefined` | - | The textual form of the color. Will have a value even if the color is invalid. `undefined` if the `value` attribute is not set. |
```
LeaVerou, line 52: That is incorrect, if the value attribute is not set you will still have a `value` property, with the current value.

DmitrySharabin: Oops. I experimented with the dev tools to check whether the `value` property was defined. I targeted the corresponding element and then tried `$0.value` in the console to see the result. Looks like it doesn't work this way, doesn't it?

LeaVerou: Oh `<color-inline>` doesn't use Props, so it doesn't follow the same standards. I'd consider it a bug.

LeaVerou (approved):

Issue #15, opened 2024-05-06
[color-gamut] Changing `Chroma increments` parameter in the Demo hangs the page


DmitrySharabin:
In all browsers.

[The Demo](https://arc.net/l/quote/rcepqqyf)

DmitrySharabin:
This happens because of an infinite loop we might get when `Chroma increments` becomes equal to `0`. This means that `step` (L58) becomes equal to `0`. And the following loop (L66-68) becomes infinite:

```js
for (let c = c_range.min; c<= c_range.max; c+=step) {
	colors.push(`oklch(${l.value}% ${c.toLocaleString("en")}% ${h.value})`);
}
```

The fix is rather simple (L52):

```diff
- <input type=number id=c_step value="0.2" min="0">
+ <input type=number id=c_step value="0.2" min="0.1" step="0.1">
```
We might want to adjust `step`, but `min` should never be equal to `0`.

PR #61, opened 2024-05-22
`<color-gamut>` refactor

LeaVerou:


src/color-gamut/color-gamut.js
```diff
@@ -10,162 +11,111 @@ const Self = class ColorGamut extends HTMLElement {
...
+				if (!gamuts) {
+					return [];
+				}
+
+				if (typeof gamuts === "string") {
+					gamuts = gamuts.trim().split(/\s*,\s*/);
+				}
+				else if (!Array.isArray(gamuts) && typeof gamuts === "object") {
+					// Object
+					return Object.entries(gamuts).map(([id, label]) => {id, label});
```
DmitrySharabin, line 73:
```suggestion
					return Object.entries(gamuts).map(([id, label]) => ({id, label}));
```

DmitrySharabin: Do I get it correctly that we expect to get an array of objects?

src/color-gamut/color-gamut.js
```diff
@@ -10,162 +11,111 @@ const Self = class ColorGamut extends HTMLElement {
...
+
+				let ret = gamuts.map((gamut, level) => {
+					if (gamut?.id && "label" in gamut) {
+						// Already in the correct format
+						return gamut;
+					}
+
+					gamut = gamut.trim().split(/\s*:\s*/);
+					let id = gamut[0];
+					let label = gamut[1] ?? Color.spaces[gamut]?.name ?? id;
```
DmitrySharabin, line 84:
```suggestion
					let label = gamut[1] ?? Color.spaces[id]?.name ?? id;
```

src/color-gamut/color-gamut.js
```diff
@@ -10,162 +11,111 @@ const Self = class ColorGamut extends HTMLElement {
...
-		this.#setGamut(gamut);
-	}
-
-	#setGamut (gamut) {
-		let oldGamut = this.#gamut;
-
-		if (gamut === oldGamut) {
-			return;
+	propChangedCallback ({name, prop, detail: change}) {
+		if (name === "gamuts") {
```
DmitrySharabin, line 29:
```suggestion
		this.#label.textContent = this.gamutLabel;

		if (name === "gamuts") {
```

src/color-gamut/color-gamut.js
```diff
@@ -10,162 +11,111 @@ const Self = class ColorGamut extends HTMLElement {
...
-	static observedAttributes = ["gamuts"];
-
-	attributeChangedCallback (name, oldValue, newValue) {
-		if (!name && this.hasAttribute("gamuts") || name === "gamuts") {
-			newValue ??= this.getAttribute("gamuts");
-
-			if (oldValue !== newValue) {
-				this.gamuts = this.constructor.parseGamuts(newValue ?? this.constructor.defaultGamuts);
-				this.#render();
+		if (name === "gamut") {
```
DmitrySharabin, line 33:
```suggestion
		if (name === "gamut") {
			this.#label.textContent = this.gamutLabel;

```

src/color-gamut/README.md
```diff
@@ -7,42 +7,20 @@ Gamut indicator. Used internally by `<color-swatch>`
...
+```html
+<color-gamut color="poop"></color-gamut>
+```
 
 Dynamic:
-```js
-colorGamutElement.color = colorValue;
+```html
+<color-gamut id="cg_1" color="red"></color-gamut>
+<script>cg_1.color = "oklch(50% 0.5 180)";</script>
```
DmitrySharabin, line 21:
```suggestion
<script>setTimeout(() => { cg_1.color = "oklch(50% 0.5 180)" }, 0);</script>
```

DmitrySharabin: We are facing a race condition with this demo so that I would propose something like this.

src/color-gamut/color-gamut.js
```diff
@@ -10,162 +11,111 @@ const Self = class ColorGamut extends HTMLElement {
...
+						id: "none",
+						get label () {
+							return ret[this.level - 1].label + "+";
+						},
+						level: ret.length,
+					});
+				}
+
+				return ret;
+			}
```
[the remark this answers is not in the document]
DmitrySharabin: Other components will break if we don't stringify gamuts.

src/color-gamut/color-gamut.js
```diff
@@ -10,162 +11,111 @@ const Self = class ColorGamut extends HTMLElement {
...
-			newValue ??= this.getAttribute("gamuts");
-
-			if (oldValue !== newValue) {
-				this.gamuts = this.constructor.parseGamuts(newValue ?? this.constructor.defaultGamuts);
-				this.#render();
+		if (name === "gamut") {
+			if (this.gamutInfo) {
+				this.style.setProperty("--gamut-level", this.gamutInfo.level);
+				this.style.setProperty("--gamut-label", this.gamutInfo.label);
+				this.style.setProperty("--gamut-id", `"${ this.gamutInfo.id }"`);
```
DmitrySharabin, line 37: I would also suggest storing `this.gamutInfo` in a variable so that the corresponding getter is not invoked four times in a row.

src/color-gamut/color-gamut.js
LeaVerou: Nope, I’m doing it with CSS generated content now so this shouldn't be needed

src/color-gamut/color-gamut.js
LeaVerou: Oh yes! But this is not new code, how did it work before? So weird.

src/color-gamut/color-gamut.js
DmitrySharabin: Oh, right. Forgot it. 😅

src/color-gamut/README.md
LeaVerou: Authors should not have to care whether the custom element has been defined or not, props should automatically take care of using any values already set on their properties instead of just overwriting them with the defaults.

However, I tried to make a testcase for this bug and can't reproduce it: https://codepen.io/leaverou/pen/rNgLOEM

DmitrySharabin: > Authors should not have to care whether the custom element has been defined or not, props should automatically take care of using any values already set on their properties instead of just overwriting them with the defaults.

Thank you!

🤷‍♂️ Not sure what kind of bug it is (and how to reproduce it), but I can show it:

https://github.com/color-js/elements/assets/9166277/b364c7a9-fb8e-4475-941a-96c4474fac1c

src/color-gamut/color-gamut.js
```diff
@@ -10,162 +11,114 @@ const Self = class ColorGamut extends HTMLElement {
...
-		if (!name && this.hasAttribute("gamuts") || name === "gamuts") {
-			newValue ??= this.getAttribute("gamuts");
-
-			if (oldValue !== newValue) {
-				this.gamuts = this.constructor.parseGamuts(newValue ?? this.constructor.defaultGamuts);
-				this.#render();
+		if (name === "gamut") {
+			if (this.gamutInfo) {
+				this.style.setProperty("--gamut-level", this.gamutInfo.level);
+				this.style.setProperty("--gamut-label", this.gamutInfo.label);
```
DmitrySharabin, line 36:
```suggestion
				this.style.setProperty("--gamut-label", `"${ this.gamutInfo.label }"`);
```

src/color-gamut/README.md
LeaVerou: Lovely, race conditions! Anyhow, I can't even get the static one to work, we should focus on that before dealing with the dynamic version.

src/color-gamut/color-gamut.js
DmitrySharabin: That's cool! Then, I proposed the correct fix below. 👇

src/color-gamut/README.md
DmitrySharabin: Should be fixed now. :)

DmitrySharabin:
I couldn't suggest a fix for the demo here (since it's not in the PR), so I pushed it directly. Now, the demo is fixed.

image

DmitrySharabin:
I would also suggest this change (mind `min` and `step`) to the demo (L52) to close #15:

```js
<p><label>Chroma increments: <input type=number id=c_step value="0.2" min="0.1" step="0.1">%</label>
```

LeaVerou:
> I would also suggest this change (mind `min` and `step`) to the demo (L52) to close #15:
> 
> ```js
> <p><label>Chroma increments: <input type=number id=c_step value="0.2" min="0.1" step="0.1">%</label>
> ```

What change is this? From what to what?

DmitrySharabin:
> What change is this? From what to what?

The demo hangs the page now (see #15). This happens because of an infinite loop we might get when `Chroma increments` becomes equal to `0`. This means that `step` (L58) becomes equal to `0`. And the following loop (L66-68) becomes infinite:

```js
for (let c = c_range.min; c<= c_range.max; c+=step) {
	colors.push(`oklch(${l.value}% ${c.toLocaleString("en")}% ${h.value})`);
}
```

The fix is rather simple (L52):

```diff
- <input type=number id=c_step value="0.2" min="0">
+ <input type=number id=c_step value="0.2" min="0.1" step="0.1">
```
We might want to adjust `step`, but `min` should never be equal to `0`.

PR #63, opened 2024-05-22
Add alpha channel support to `<channel-slider>` and `<color-picker>`

DmitrySharabin:


src/channel-slider/channel-slider.js
```diff
@@ -178,7 +192,8 @@ export default class ChannelSlider extends NudeElement {
 		defaultValue: {
 			type: Number,
 			default () {
-				return this.defaultColor.get(this.channel);
+				let color = this.defaultColor;
+				return this.channel === "alpha" ? color.alpha : color.get(this.channel);
```
LeaVerou, line 196: Does `color.get("alpha")` not work? I think that's a reported Color.js bug, it may be worth it to fix it there

src/channel-slider/channel-slider.js
```diff
@@ -140,6 +147,13 @@ export default class ChannelSlider extends NudeElement {
 		},
 		channelSpec: {
 			get () {
+				if (this.channel === "alpha") {
+					return {
+						name: "Alpha",
+						range: [0, 1],
```
LeaVerou, line 153: Ideally we'd want to expose a 0-100% percentage to users

src/channel-slider/channel-slider.js
```diff
@@ -42,12 +42,19 @@ export default class ChannelSlider extends NudeElement {
...
-			color.set(this.channel, value);
+
+		if (this.channel === "alpha") {
+			color.alpha = value;
 		}
-		catch (e) {
-			console.warn(e);
+		else {
+			try {
+				color.set(this.channel, value);
```
LeaVerou, line 51: Same as below, does `color.set("alpha", value)` not work? It should.

src/channel-slider/channel-slider.js
DmitrySharabin: Then, I think we can get rid of `range`; the default `[0, 100]` will be used. But that means we need to adjust the value on every change, right?

LeaVerou (changes requested):
How does it behave in Color.js < v0.5.0? It should not error if alpha is not used, and should print out some kind of warning that Using alpha requires v0.6.0 or higher.

DmitrySharabin:
> How does it behave in Color.js < v0.5.0? It should not error if alpha is not used, and should print out some kind of warning that Using alpha requires v0.6.0 or higher.

It doesn't do it now. But will. :)

DmitrySharabin:
This PR is outdated. I'm gonna close it in favor of #173 and the one for `<color-picker>` that follows.

PR #65, opened 2024-05-27
`step` should be a positive number, fixes #15

### LINKED ISSUE #15: [color-gamut] Changing `Chroma increments` parameter in the Demo hangs the page

DmitrySharabin:


LeaVerou (approved):
I would have probably done `step = Math.max(0, step)` but this is fine too.

DmitrySharabin:
There is a chance that it might become `0`, and we’ll get an infinite loop again. 🙂

PR #68, opened 2024-05-28
[channel-slider] Add docs on getters

DmitrySharabin:


src/channel-slider/README.md
```diff
@@ -126,8 +126,19 @@ All attributes are reactive:
...
+
+### Getters
+
+These properties are read-only.
+
+| Property | Type | Description |
+|----------|------|-------------|
+| `minColor` | `Color` | The color corresponding to the minimum value of the slider. |
+| `maxColor` | `Color` | The color corresponding to the maximum value of the slider. |
+| `stops` | `Array<Color>` | The array of color stops corresponding to the slider progress of `0%`, `25%`, `50%`, `75%`, and `100%` respectively. |
```
LeaVerou, line 138: I don’t think we should tie it to specific percentages, it depends on the case and browser support.

src/channel-slider/README.md
```diff
@@ -126,8 +126,19 @@ All attributes are reactive:
...
+| `progress` | `number` | The slider progress calculated on its current value (e.g. `0.42`). |
```
LeaVerou, line 139: As a number from 0 to 1.

src/channel-slider/README.md
```diff
@@ -126,8 +126,19 @@ All attributes are reactive:
...
+| `channelName` | `string` | The label of the slider (e.g. `Hue`). |
```
LeaVerou, line 140: The name of the channel (e.g. Hue or Alpha). We should also have an example with alpha.

LeaVerou (changes requested):

src/channel-slider/README.md
```diff
@@ -126,8 +126,19 @@ All attributes are reactive:
...
+### Getters
+
+These properties are read-only.
+
+| Property | Type | Description |
+|----------|------|-------------|
+| `minColor` | `Color` | The color corresponding to the minimum value of the slider. |
+| `maxColor` | `Color` | The color corresponding to the maximum value of the slider. |
+| `stops` | `Array<Color>` | The array of the slider color stops with `minColor` and `maxColor` as its first and last elements, respectively. |
+| `progress` | `number` | The slider progress, as a number between `0` and `1`, calculated on its current value (e.g. `0.42`). |
```
LeaVerou, line 139:
```suggestion
| `progress` | `number` | The slider value converted to a 0-1 number with `0` corresponding to the min of the range and `1` to the max. |
```

src/channel-slider/README.md
```diff
@@ -126,8 +126,19 @@ All attributes are reactive:
...
+
+### Getters
+
+These properties are read-only.
+
+| Property | Type | Description |
+|----------|------|-------------|
+| `minColor` | `Color` | The color corresponding to the minimum value of the slider. |
+| `maxColor` | `Color` | The color corresponding to the maximum value of the slider. |
+| `stops` | `Array<Color>` | The array of the slider color stops with `minColor` and `maxColor` as its first and last elements, respectively. |
```
LeaVerou, line 138:
```suggestion
| `stops` | `Array<Color>` | The array of the slider color stops used for rendering. Unsupported color spaces or angular channels (hues) will have more color stops, while other channels may have as little as two: `minColor` and `maxColor`. |
```

LeaVerou (approved):

src/channel-slider/README.md
DmitrySharabin: We should probably update the corresponding description for `<color-slider>`, too.

PR #88, opened 2024-06-05
[color-slider] Correctly handle spinner properties when showing progress

### LINKED ISSUE #16: `tooltip="progress"` on `<color-slider>` (and other related elements) inherits `min`, `max`, and `step` from the slider

DmitrySharabin:
Fixes #16

### DIFF
--- a/src/color-slider/color-slider.js
+++ b/src/color-slider/color-slider.js
@@ -132,9 +132,9 @@ const Self = class ColorSlider extends NudeElement {
 			}
 		}
 		else if (name === "tooltip") {
-			if (change.value === "progress") {
+			if (change.value !== undefined) {
 				["min", "max", "step", "value"].forEach(name => {
-					this._el.spinner[name] = values[name];
+					this._el.spinner[name] = change.value === "progress" ? values[name] : this[name];
 				});
 			}
 		}
--- a/src/color-slider/color-slider.js
+++ b/src/color-slider/color-slider.js
@@ -140,7 +140,10 @@ const Self = class ColorSlider extends NudeElement {
 			if (change.value !== undefined) {
 				let values = this;
 				if (change.value === "progress") {
-					values = { min: 1, max: 100, step: 1, value: +(this.progress * 100).toPrecision(4) };
+					values = {
+						min: 1, max: 100, step: 1,
+						value: +(this.progress * 100).toPrecision(4),
+					};
 				}
 
 				["min", "max", "step", "value"].forEach(name => {
--- a/src/color-slider/color-slider.js
+++ b/src/color-slider/color-slider.js
@@ -60,13 +60,14 @@ const Self = class ColorSlider extends NudeElement {
 	}
 
 	propChangedCallback ({name, prop, detail: change}) {
+		// Spinner values when tooltip is "progress"
+		let values = { min: 1, max: 100, step: 1, value: +(this.progress * 100).toPrecision(4) };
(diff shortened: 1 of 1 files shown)


src/color-slider/color-slider.js
```diff
@@ -130,6 +131,13 @@ const Self = class ColorSlider extends NudeElement {
 				this._el.spinner.style.setProperty("--value-length", valueStr.length);
 			}
 		}
+		else if (name === "tooltip") {
+			if (change.value === "progress") {
```
LeaVerou, line 135: Don’t we still want to set these (to different values) even if the value is not progress?

DmitrySharabin: Yes, we do. I also thought about it today. I will fix it. Thanks.

DmitrySharabin: Done!

src/color-slider/color-slider.js
```diff
@@ -60,13 +60,14 @@ const Self = class ColorSlider extends NudeElement {
 	}
 
 	propChangedCallback ({name, prop, detail: change}) {
+		// Spinner values when tooltip is "progress"
+		let values = { min: 1, max: 100, step: 1, value: +(this.progress * 100).toPrecision(4) };
```
LeaVerou, line 64: We should not calculate this no matter what, but inside relevant conditionals.

LeaVerou (changes requested):

src/color-slider/color-slider.js
DmitrySharabin: Agreed. My motivation was to make the code a bit DRYer. Fixed.

src/color-slider/color-slider.js
```diff
@@ -64,8 +64,15 @@ const Self = class ColorSlider extends NudeElement {
...
-			if (this.tooltip === "progress" && (name === "value" || name === "defaultValue")) {
-				value = +(this.progress * 100).toPrecision(4);
+			if (this.tooltip === "progress") {
+				if (name === "value" || name === "defaultValue") {
+					value = +(this.progress * 100).toPrecision(4);
+				}
+				else {
+					// Spinner values when tooltip is "progress"
+					let values = { min: 1, max: 100, step: 1 };
+					value = values[name];
```
LeaVerou, line 74:
```suggestion
					value = ({ min: 1, max: 100, step: 1 })[name];
```

(just a suggestion, only accept if you find it more readable)

src/color-slider/color-slider.js
```diff
@@ -130,6 +137,18 @@ const Self = class ColorSlider extends NudeElement {
 				this._el.spinner.style.setProperty("--value-length", valueStr.length);
 			}
 		}
+		else if (name === "tooltip") {
+			if (change.value !== undefined) {
+				let values = this;
+				if (change.value === "progress") {
+					values = { min: 1, max: 100, step: 1, value: +(this.progress * 100).toPrecision(4) };
```
LeaVerou, line 144: Nit, optional: I would break this into more lines

LeaVerou (approved):

Issue #85, opened 2024-06-04
[color-swatch] Regression: If `color` is set programmatically, `value` doesn't update


DmitrySharabin:
[Testcase](https://codepen.io/dmitrysharabin/pen/WNBjLYy)

```html
<color-swatch id="dynamic_static">oklch(70% 0.25 138)</color-swatch>
<button onclick='dynamic_static.color = "oklch(60% 0.15 0)"'>Change color</button>
```

DmitrySharabin:
One more thing to note. If `<color-swatch>` is editable after the color is set programmatically, changing the value in `<input>` no longer updates the color. The link between `value` and `color` is broken after the color change.

DmitrySharabin:
> The link between `value` and `color` is broken after the color change.

OK, this is why it happens (at least, as far as I can see).

`value` is the default prop for `color`:  https://github.com/color-js/elements/blob/92cb2e0f8a7beb09eb4f23f807ff2f03b58e82c9/src/color-swatch/color-swatch.js#L134
Every time we update `value`, `color`, as a dependent of `value`, is also updated. This happens because [`dependsOn()`](https://github.com/nudeui/element/blob/924a27fb979f1b38e0b3e6b8897159759350ca5e/src/props/Prop.js#L290-L301)
returns `true`.

But why does it return `true` in our case? Because of [this line](https://github.com/nudeui/element/blob/924a27fb979f1b38e0b3e6b8897159759350ca5e/src/props/Prop.js#L300):
```js
this.defaultProp === prop && element.props[this.name] === undefined
```
`value` is the `color` default prop and `element.props["color"] === undefined` (since we haven't set it programmatically yet).

When we set `color` programmatically, we [set the corresponding prop on the element](https://github.com/nudeui/element/blob/924a27fb979f1b38e0b3e6b8897159759350ca5e/src/props/Prop.js#L196). This means `element.props["color"]` is not `undefined` anymore. In its turn, it means that `dependsOn()` will return `false` when the `value` prop updates, and the dependent (`color`) won't be updated—changes to `value` won't be propagated to `color`.

I wonder, what should be done in this situation?

- `value` should not be the default prop for `color` and `color`, in its turn, should get its default value the other way?
- When setting a value to a prop that has a default prop, should we update the default prop accordingly and not set `element.props[...]` for this prop to preserve the link between the prop and its default prop?
- We should update the logic of `dependsOn()` to cover such cases?

DmitrySharabin:
[...]

PR #91, opened 2024-06-06
[color-picker] If no color is provided, use the default one

DmitrySharabin:
This PR also fixes this Codepen: https://codepen.io/leaverou/pen/wvbzovK

### DIFF
--- a/src/color-picker/color-picker.js
+++ b/src/color-picker/color-picker.js
@@ -80,13 +80,20 @@ const Self = class ColorPicker extends NudeElement {
 		}
 
 		if (name === "color") {
-			for (let slider of this._el.sliders.children) {
-				slider.color = this.color;
-			}
+			if (this.color) {
+				for (let slider of this._el.sliders.children) {
+					slider.color = this.color;
+				}
 
-			if (this.color && (!this._el.swatch.color || !this.color.equals(this._el.swatch.color))) {
-				// Avoid typing e.g. "red" and having it replaced with "rgb(100% 0% 0%)" under your caret
-				prop.applyChange(this._el.swatch, { ...change, source: "property" });
+				if (this._el.swatch.color || !this.color.equals(this._el.swatch.color)) {
+					// Avoid typing e.g. "red" and having it replaced with "rgb(100% 0% 0%)" under your caret
+					prop.applyChange(this._el.swatch, { ...change, source: "property" });
+				}
+			}
+			else {
+				// No color is provided; use to the default one.
+				// We should do this to propagate the default color to all the underlying elements: the sliders and swatch.
+				this.color = this.defaultColor;
 			}
 		}
 	}


LeaVerou (changes requested):
Hmmm not sure I like setting `color` when it's just the default. Then we lose track of whether this has been user set or just the default value. What if we instead change the condition to `name === "color" || name === "defaultColor"`?

DmitrySharabin:
> Hmmm not sure I like setting `color` when it's just the default. Then we lose track of whether this has been user set or just the default value. What if we instead change the condition to `name === "color" || name === "defaultColor"`?

This works, too. One thing that stopped me (when I tried it yesterday) from using this approach is that the `colorchange` event is not fired after the component is mounted in that case. This is expected from my perspective, but is this what the users expect?

LeaVerou:
> > Hmmm not sure I like setting `color` when it's just the default. Then we lose track of whether this has been user set or just the default value. What if we instead change the condition to `name === "color" || name === "defaultColor"`?
> 
> This works, too. One thing that stopped me (when I tried it yesterday) from using this approach is that the `colorchange` event is not fired after the component is mounted in that case. This is expected from my perspective, but is this what the users expect?

That sounds like a separate bug. If `color` has no value and its default value changes, `propchange` (and thus `colorchange`) *should* fire. Instead of working around our own bugs, we should be fixing them!

src/color-picker/color-picker.js
```diff
@@ -79,14 +79,16 @@ const Self = class ColorPicker extends NudeElement {
 			}
 		}
 
-		if (name === "color") {
+		if (name === "color" || name === "defaultColor") {
+			let color = this.color ?? this.defaultColor;
```
LeaVerou, line 83: `this.color` should be equal to `this.defaultColor` anyway. Is it not?

DmitrySharabin:
> That sounds like a separate bug. If color has no value and its default value changes, `propchange` (and thus `colorchange`) should fire. Instead of working around our own bugs, we should be fixing them!

> `this.color` should be equal to `this.defaultColor` anyway if it has no value. Is it not?

OK, here is the current state after I took another step toward solidifying my knowledge of `NudeElement` and color elements props.

For now, `defaultColor` _is not_ the default prop for `color`:  https://github.com/color-js/elements/blob/13a7898e371975237b497192dd6eea0b3725bfc7/src/color-picker/color-picker.js#L131-L136

That's why we might stumble upon a situation when `this.color` is `undefined`. Until we implement https://github.com/nudeui/element/issues/24, we can't simply use `defaultProp: "defaultColor"` for `color`; we have to use a getter and setter instead (see https://github.com/color-js/elements/issues/85#issuecomment-2149710030 as an example of why it happens).

With this PR, we fix the bugs mentioned above by Lea instead of working around them.

Another thing is that when we set `space` (whether specified or not, but especially if it wasn't), we should update the color of underlying sliders and _not build a new color_ based on the sliders' current values. With this change, [the color picker from docs](https://elements.colorjs.io/src/color-picker/#dynamic) and [the one from `Color.js`](https://apps.colorjs.io/picker) work _identically_ (for cases when `alpha` is equal to 1). And this is definitely a good thing! 🙂

So, let's get another iteration on this started.

src/color-picker/color-picker.js
```diff
@@ -130,8 +129,11 @@ const Self = class ColorPicker extends NudeElement {
 
 		color: {
 			type: Color,
+			get () {
+				return this.defaultColor;
+			},
 			set (value) {
-				this.defaultColor = new Color(value).to(this.space);
+				this.defaultColor = value;
```
DmitrySharabin, line 136: We don't need `new Color(value).to(this.space)` anymore because `defaultColor.convert()` has our back.

LeaVerou: Wait, is `color` and `defaultColor` really the same thing? If so, why do we even have two props?

DmitrySharabin: We don't need two, indeed. IDK why I kept them (even though I was thinking of merging them). 🤷‍♂️
Now, I merged them.

LeaVerou:
Pretty sure there was a reason they were separate though. 🤔

LeaVerou:
I think one of them was meant to represent the color provided in the HTML (or the color space midpoint, if none was provided), and the other one the *current* color. I.e. `defaultColor` was not meant to change.

LeaVerou:
> > I think one of them was meant to represent the color provided in the HTML (or the color space midpoint, if none was provided), and the other one the _current_ color. I.e. `defaultColor` was not meant to change.
> 
> Makes sense!
> 
> It means that we could have something like this, right?
> 
> ```js
> defaultColor: {
> 	type: Color,
> 	convert (color) { },
> 	default () { },
> 	reflect: {
> 		from: "color",
> 	},
> },
> 
> color: {
> 	type: Color,
> 	defaultProp: "defaultColor",
> 	reflect: false,
> },
> ```
I belive so, yes.
 
> We might want to wait for changes on Nude Element to make everything right or use changes in this PR as a workaround and fix everything properly after Nude Element is ready. What do you think?

What changes?

DmitrySharabin:
> What changes? This is not the same as rawProp/parsedProp.

Right, it has nothing to do with the issue I'm trying to solve. 🤦‍♂️

Splitting `color` and `defaultColor` actually works really fine. What was I thinking of earlier? It's such a straightforward thing to try. 🤷‍♂️

src/color-picker/color-picker.js
```diff
@@ -84,9 +83,13 @@ const Self = class ColorPicker extends NudeElement {
 				slider.color = this.color;
 			}
 
-			if (this.color && (!this._el.swatch.color || !this.color.equals(this._el.swatch.color))) {
+			if (!this._el.swatch.color || !this.color.equals(this._el.swatch.color)) {
 				// Avoid typing e.g. "red" and having it replaced with "rgb(100% 0% 0%)" under your caret
-				prop.applyChange(this._el.swatch, { ...change, source: "property" });
+				if (change.source === "attribute") {
+					// <color-swatch> doesn't have a color attribute, so we must change the corresponding property
```
LeaVerou, line 89: it …does? https://elements.colorjs.io/src/color-swatch/#the-color-attribute

LeaVerou (approved):
LGTM, though see comment around that `color` attribute.

src/color-picker/color-picker.js
DmitrySharabin: Actually, this is not the case after [we fixed `<color-swatch>`](https://github.com/color-js/elements/issues/85) (by replacing `defaultProp` with a getter and setter). I'll remove this check with the comment entirety. Removing the code and making our components even lighter is so cool. 🤩

PR #93, opened 2024-06-06
[color-swatch][color-scale] First stab at implementing the coords prop

DmitrySharabin:
It's a very basic implementation, so we could have something to start with. 🙂
It looks like this:

SCR-20240607-bwjc

When defining the type of `coords`, I used a workaround because of [the issue](https://github.com/nudeui/element/issues/27) I faced.

I wonder whether we should do something like `+(value).toPrecision(4)` to make the output look nicer (or is it OK to see that many digits?). Is there a way to specify precision in `Color.js` we could use?

**UPDATE**

Here is the code that produces the image above:

```html
<color-swatch coords="oklch.l, oklch.c, oklch.h" size="large">oklch(80% 50% 70)</color-swatch>

<div id="future_swatch_container"></div>
<script>
	let swatch = document.createElement("color-swatch");
	swatch.color = "oklch(65% 0.15 210)";
	swatch.setAttribute("size", "large");
	swatch.textContent = "Turquoise";
	swatch.coords = "L: oklch.l, C: oklch.c, H: oklch.h";
	future_swatch_container.append(swatch);
</script>

<color-scale colors="#e3fafc, #0b7285" steps="4" space="oklch" coords="L: oklch.l, C: oklch.c"></color-scale>

<color-scale space="oklch" colors="
	Gray 50: #f9fafb,
	Gray 100: #f3f4f6,
	Gray 200: #e5e7eb,
	Gray 300: #d1d5db,
	Gray 400: #9ca3af
" coords="srgb.r"></color-scale>
```

### DIFF
--- a/src/color-scale/color-scale.js
+++ b/src/color-scale/color-scale.js
@@ -66,8 +66,10 @@ const Self = class ColorScale extends NudeElement {
 			}
 
 			swatch.color = color;
-			swatch.coords = this.coords;
 			swatch.textContent = name;
+			if (this.coords) {
+				swatch.coords = this.coords;
+			}
 			i++;
 		}
 
--- a/src/color-scale/color-scale.js
+++ b/src/color-scale/color-scale.js
@@ -148,7 +148,7 @@ const Self = class ColorScale extends NudeElement {
 
 				return colors;
 			},
-			dependencies: ["colors", "steps", "space", "coords"],
+			additionalDependencies: ["coords"],
 		},
 		coords: {},
 	};
--- a/src/color-scale/color-scale.js
+++ b/src/color-scale/color-scale.js
@@ -66,6 +66,7 @@ const Self = class ColorScale extends NudeElement {
 			}
 
 			swatch.color = color;
+			swatch.coords = this.coords;
 			swatch.textContent = name;
 			i++;
 		}
@@ -144,7 +145,11 @@ const Self = class ColorScale extends NudeElement {
 				}
 
 				return colors;
-			}
+			},
+			dependencies: ["colors", "steps", "space", "coords"],
+		},
+		coords: {
+			type: String,
 		},
 	};
 }
--- a/src/color-scale/color-scale.js
+++ b/src/color-scale/color-scale.js
@@ -150,9 +150,7 @@ const Self = class ColorScale extends NudeElement {
 			},
 			dependencies: ["colors", "steps", "space", "coords"],
 		},
-		coords: {
-			type: String,
-		},
+		coords: {},
 	};
 }
 
--- a/src/color-swatch/color-swatch.css
+++ b/src/color-swatch/color-swatch.css
@@ -74,6 +74,10 @@ slot {
 			font-variant-numeric: tabular-nums;
 		}
 	}
+
+	&:empty {
+		display: none;
+	}
 }
 
 #wrapper {
--- a/src/color-swatch/color-swatch.css
+++ b/src/color-swatch/color-swatch.css
@@ -49,6 +49,23 @@ slot {
 	}
 }
 
+#coords {
+	margin: 0;
+	display: grid;
+	grid-template-columns: max-content auto;
+	gap: .3em .7em;
+
+	.coord {
+		display: contents;
+
+		dd {
+			margin: 0;
+			font-weight: bold;
+			font-variant-numeric: tabular-nums;
+		}
+	}
+}
+
 #wrapper {
 	display: flex;
 	flex-flow: inherit;
--- a/src/color-swatch/color-swatch.css
+++ b/src/color-swatch/color-swatch.css
@@ -51,12 +51,22 @@ slot {
 
 #coords {
 	margin: 0;
-	display: grid;
-	grid-template-columns: max-content auto;
-	gap: .3em .7em;
+	display: inline-flex;
+	gap: .5em;
+
+	&:is(:host([size="large"]) &) {
+		display: grid;
+		grid-template-columns: max-content auto;
+		gap: .3em .7em;
+
+		.coord {
+			display: contents;
+		}
+	}
 
 	.coord {
-		display: contents;
+		display: flex;
+		gap: .2em;
 
 		dd {
 			margin: 0;
--- a/src/color-swatch/color-swatch.css
+++ b/src/color-swatch/color-swatch.css
@@ -49,7 +49,7 @@ slot {
 	}
 }
 
-#coords {
+[part="coords"] {
 	margin: 0;
 	display: inline-flex;
 	gap: .5em;
@@ -73,10 +73,6 @@ slot {
 			font-weight: bold;
 			font-variant-numeric: tabular-nums;
 		}
-	}
-
-	&:empty {
-		display: none;
 	}
 }
 
--- a/src/color-swatch/color-swatch.js
+++ b/src/color-swatch/color-swatch.js
@@ -24,6 +24,7 @@ const Self = class ColorSwatch extends NudeElement {
 				<div part="color-wrapper">
 					<slot></slot>
 				</div>
+				<dl id="coords" part="coords"></dl>
 				<slot name="after"></slot>
 			</div>
 		`;
@@ -31,6 +32,7 @@ const Self = class ColorSwatch extends NudeElement {
 		this._el = {};
 		this._el.wrapper = this.shadowRoot.querySelector("#wrapper");
 		this._el.colorWrapper = this.shadowRoot.querySelector("[part=color-wrapper]");
+		this._el.coords = this.shadowRoot.querySelector("#coords");
 		this._el.slot = this.shadowRoot.querySelector("slot:not([name])");
 
 		this.#updateStatic();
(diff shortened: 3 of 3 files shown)


src/color-scale/color-scale.js
```diff
@@ -144,7 +147,11 @@ const Self = class ColorScale extends NudeElement {
 				}
 
 				return colors;
-			}
+			},
+			dependencies: ["colors", "steps", "space", "coords"],
```
LeaVerou, line 151: We should have an `additionalDependencies` property rather than having to redefine all of them. Can you file it?

src/color-scale/color-scale.js
```diff
@@ -144,7 +147,11 @@ const Self = class ColorScale extends NudeElement {
...
+		},
+		coords: {
+			type: String,
```
LeaVerou, line 154: Why is this `type: String`? At the very least it should be an array or object.

src/color-swatch/color-swatch.js
```diff
@@ -109,6 +111,26 @@ const Self = class ColorSwatch extends NudeElement {
 			this.style.setProperty("--color", colorString);
 		}
 
+		if (name === "coords") {
+			if (this.color) {
+				this._el.coords?.replaceChildren(); // remove all children
```
LeaVerou, line 116: Nit: You don't need `replaceChildren()` to remove all children, `el.textContent = ""` works and is more common.
Why the optional chaining? Is there any chance this runs before `this._el.coords` is defined? I thought we fixed that?

src/color-swatch/color-swatch.js
```diff
@@ -109,6 +111,26 @@ const Self = class ColorSwatch extends NudeElement {
...
+					let [label, properties] = Object.entries(coord)[0];
+					let [space, channel] = properties.split(".");
+					let value;
+
+					if (!channel) {
+						value = this.color[space];
+					}
+					else {
+						value = this.color[space][channel];
+					}
```
LeaVerou, line 128:
```suggestion
					let value = this.color[space];

					if (channel) {
						value = value[channel];
					}
```

src/color-swatch/color-swatch.js
```diff
@@ -109,6 +111,26 @@ const Self = class ColorSwatch extends NudeElement {
...
+
+					this._el.coords.insertAdjacentHTML("beforeend", `<div class="coord"><dt>${ label }</dt><dd>${ value }</dd></div>`);
```
LeaVerou, line 130: Let's minimize DOM operations by building this string and then setting the element’s entire `innerHTML` to it. You can build an array with each entry and then set innerHTML to that array with `.join("\n")`

src/color-scale/color-scale.js
LeaVerou: Oh I see, because it just passes itself verbatim to `<color-swatch>`. You don't need `type: String` then, just don't include a type at all. In the future we'll probably need to parse it in `<color-scale>` though, since it could be useful for `<color-scale>` to have summary stats on these coords. But for now just passing it through is ok.

src/color-swatch/color-swatch.js
```diff
@@ -143,6 +165,28 @@ const Self = class ColorSwatch extends NudeElement {
...
+				values: {
+					is: Object,
+					defaultKey: (value, i) => {
+						let [space, channel] = value.split(".");
+						if (!channel) {
+							channel = space;
+							space = this.color?.space.id;
+						}
+
+						return Color.Space.get(space)?.coords[channel]?.name ?? channel;
```
LeaVerou, line 180: Please use `Color.Space.resolveCoord()` which exists exactly for this purpose: https://github.com/color-js/color.js/blob/main/src/ColorSpace.js#L370

src/color-swatch/color-swatch.js
```diff
@@ -143,6 +165,28 @@ const Self = class ColorSwatch extends NudeElement {
...
+					},
+				},
+			},
+			default: [],
+			reflect: {
+				from: true,
+			},
+			dependencies: ["color"],
```
LeaVerou, line 188: So you found a bug in Nude Element: that `defaultKey()` (and `defaultValue()`) isn't taken into account when calculating dependencies. Don't just work around bugs, file them so we can fix them!

LeaVerou (changes requested):

src/color-scale/color-scale.js
DmitrySharabin: Done!

https://github.com/nudeui/element/issues/28

src/color-swatch/color-swatch.js
DmitrySharabin: > Nit: You don't need `replaceChildren()` to remove all children, `el.textContent = ""` works and is more common.

This was my first guess. But I wasn't sure it was a common practice. Now I know. Thanks!

> Why the optional chaining? Is there any chance this runs before `this._el.coords` is defined? I thought we fixed that?

It's. My bad. Will remove it.

src/color-swatch/color-swatch.js
DmitrySharabin: With great pleasure! TIL. Thanks for sharing it with me.

src/color-swatch/color-swatch.css
```diff
@@ -49,6 +49,33 @@ slot {
 	}
 }
 
+#coords {
+	margin: 0;
+	display: inline-flex;
+	gap: .5em;
```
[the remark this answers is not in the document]
DmitrySharabin: Agreed!

src/color-swatch/color-swatch.js
DmitrySharabin: More than agreed. 🙂

src/color-swatch/color-swatch.js
DmitrySharabin: You think about me too well! 😂 I didn't work around the bug; I knew we infer dependencies from `default`, `get`, and `convert`. And for me, it's just like that. So, I just worked with the current implementation. I'm happy that thanks to my “discovery,” we'll make `NudeElement` even better, though. 🙂

DmitrySharabin:
I addressed all your feedback. Thank you so much! It should be ready now for another iteration.

src/color-swatch/color-swatch.css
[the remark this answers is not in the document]
LeaVerou: Actually, how hard would it be to not have the `<dl>` at all in that case?

src/color-swatch/color-swatch.js
```diff
@@ -109,6 +111,27 @@ const Self = class ColorSwatch extends NudeElement {
...
 		}
 
+		if (name === "coords") {
+			if (this.color) {
+				this._el.coords.textContent = ""; // remove all children
+
+				let coords = [];
+				for (let coord of this.coords) {
+					let [label, properties] = Object.entries(coord)[0];
+					let [space, channel] = properties.split(".");
```
LeaVerou, line 121: This is still parsing coords manually

LeaVerou (changes requested):
See comments

src/color-swatch/color-swatch.js
LeaVerou: ANY tool you use can have bugs. That includes popular libraries, frameworks, even the browser itself. Always try to report them. And here we can even fix them at the source!

src/color-swatch/color-swatch.js
DmitrySharabin: Oops. Fixed.

src/color-swatch/color-swatch.css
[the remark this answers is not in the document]
DmitrySharabin: That's a much better solution—not to have `<dl>` if unnecessary. Thank you for the idea. It should work that way from now on.

DmitrySharabin:

Before requesting your review again, I'd love to know if there is a way to output formatted values for coordinates (the way they are displayed in the parsed color). We might get a result much closer to the one you showcased in the issue:

image

LeaVerou:
No, but that's an excellent feature request for `getAll()`.

src/color-swatch/color-swatch.js
```diff
@@ -109,6 +109,32 @@ const Self = class ColorSwatch extends NudeElement {
 			this.style.setProperty("--color", colorString);
 		}
 
+		if (name === "coords") {
+			if (!this.coords.length) {
+				return;
+			}
+
+			if (!this._el.coords) {
+				this._el.colorWrapper.insertAdjacentHTML("afterend", `<dl part="coords"></dl>`);
```
LeaVerou, line 118: Nit but I'd rather only insert this once and then keep it around in `this._el.coords`:

```js
this._el.coords ??= Object.assign(document.createElement("dl"), {part: "coords"});
```

DmitrySharabin: Q: Do we want to place `this._el.coords` inside `this._el.colorWrapper`? Previously, it was its sibling.

DmitrySharabin:
> No, but that's an excellent feature request for `getAll()`.

Do you mean `Color.getAll()` in `Color.js`? Should I file an issue?

src/color-swatch/color-swatch.js
LeaVerou: No, use `.before()` instead

LeaVerou:
> > No, but that's an excellent feature request for `getAll()`.
> 
> Do you mean `Color.getAll()` in `Color.js`? Should I file an issue?

Yes

LeaVerou:
Don't worry about formatting for now, just apply precision. We'll deal with formatting in Color.js.

DmitrySharabin:
> Don't worry about formatting for now, just apply precision. We'll deal with formatting in Color.js.

Sure. Just to ensure I follow: we can do it like this: `value.toPrecision(4)` Right? Or do we have a way to get the needed precision with `Color.js` (with some feature I'm not yet aware of 🫣)?

LeaVerou:
Yup, `+value.toPrecision()`.

LeaVerou (approved):

PR #100, opened 2024-06-14
[color-chart] Experiments

DmitrySharabin:
Here are some of my experiments with `<color-chart>`. I tried to add support for an optional label and moved some additional info to the `title` of the corresponding swatch.

### DIFF
--- a/src/color-chart/README.md
+++ b/src/color-chart/README.md
@@ -25,3 +25,53 @@ By default, the other coordinate would be the index of the color in the list, bu
 	<color-scale colors="50: #e3f2fd, 100: #bbdefb, 200: #90caf9, 300: #64b5f6, 400: #42a5f5, 500: #2196f3, 600: #1e88e5, 700: #1976d2, 800: #1565c0, 900: #0d47a1"></color-scale>
 </color-chart>
 ```
+
+You can optionally specify a label for a data point by  dividing it from the other coordinate with `/` (slash):
+
+```html
+<color-chart y="oklch.l">
+	<color-scale colors="
+		Blue 50 / 50: #e3f2fd,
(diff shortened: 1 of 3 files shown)


src/color-chart/README.md
```diff
@@ -24,4 +24,54 @@ By default, the other coordinate would be the index of the color in the list, bu
...
 	<color-scale colors="50: #e3f2fd, 100: #bbdefb, 200: #90caf9, 300: #64b5f6, 400: #42a5f5, 500: #2196f3, 600: #1e88e5, 700: #1976d2, 800: #1565c0, 900: #0d47a1"></color-scale>
 </color-chart>
+```
+
+You can optionally specify a label for a data point by  dividing it from the other coordinate with `/` (slash):
+
+```html
+<color-chart y="oklch.l">
+	<color-scale colors="
+		Blue 50 / 50: #e3f2fd,
```
LeaVerou, line 34: This repetition should not be needed (and I already have code for that — though no idea if it works)

src/color-chart/color-chart-global.css
```diff
@@ -40,9 +40,12 @@ color-chart > color-scale {
 	}
 
 	&::part(color-swatch)::before {
-		content: "Hi";
+		content: var(--label);
```
LeaVerou, line 43: Don't think we really want a label, given the usual density of this kind of data, but I suppose it can't hurt? Def not a priority though, especially before we get the basic stuff working…

LeaVerou: I was planning to use that to draw the line

DmitrySharabin:
Lea did a great job with `<color-chart>`, so there is no need for this experiment at all.

PR #104, opened 2024-06-18
[color-chart] Place the Y-axis underneath a tooltip with the value

DmitrySharabin:
...when the chart container is hovered.

I tried another approach, but I couldn't get the shadow DOM boundaries to style the axis based on the state of an element in the light DOM. I also tried creating a new stacking context for the tooltip with `isolation: isolate` without luck.

Without the fix, we might end up with the following:

SCR-20240618-kukd

With the fix, we have this:

SCR-20240618-kvay

### DIFF
--- a/src/color-chart/color-chart.css
+++ b/src/color-chart/color-chart.css
@@ -73,6 +73,9 @@
 	container-name: chart;
 	container-type: inline-size;
 
+	&:hover ~ #y_axis {
+		z-index: -1;
+	}
 }
 
 #chart {


LeaVerou (changes requested):
I would avoid `z-index: -1` as it can end up making the element be hidden behind containers. What about applying `z-index: 1` to the color scales?

DmitrySharabin:
> I would avoid `z-index: -1` as it can end up making the element be hidden behind containers.

Yes, this bothers me, too.

> What about applying `z-index: 1` to the color scales?

We already do it, unfortunately, with no effect.

https://github.com/color-js/elements/blob/0661aa7a74dd129dd103968eefc89296001c98c2/src/color-chart/color-chart-global.css#L51

Moreover, the tooltip itself has a `z-index` set:

https://github.com/color-js/elements/blob/0661aa7a74dd129dd103968eefc89296001c98c2/src/color-chart/color-chart-global.css#L68

I tried changing both of them. No effect.

I'll try to find some other options, then.

DmitrySharabin:
It's not an issue anymore. 🥳 So, I'm going to close it.

PR #105, opened 2024-06-18
[color-chart] Don't show points tooltips on hovering the line

DmitrySharabin:
For now, when we hover a line between two points, the tooltip for the left one is shown (even when we are close enough to the right point of the two):

image

To my mind (and expectations), we should only show tooltips on points hover. This PR should fix it.


LeaVerou (approved):
Haha nice

PR #118, opened 2024-10-11
[color-swatch] Add the `colorInfo` prop

DmitrySharabin:
As we discussed in #96, the color info is also helpful to get programmatically. This PR addresses this. Under the hood, [the data structure suggested by @LeaVerou (Option 1)](https://github.com/color-js/elements/pull/96#issuecomment-2173355441) is used.

This PR is the first step to implementing deltas and contrast and is intended to check if I got the main idea shaped in the PR mentioned right. The existing API is the same, so it's not a breaking change. Everything still works as before.

The gist of what this PR introduces can be seen in the screenshot (`<color-swatch>` is a Lego block for color charts):

SCR-20241011-peso

### DIFF
--- a/src/color-swatch/color-swatch.js
+++ b/src/color-swatch/color-swatch.js
@@ -115,8 +115,8 @@ const Self = class ColorSwatch extends ColorElement {
 			this.style.setProperty("--color", colorString);
 		}
 
-		if (name === "info" || name === "color") {
-			if (!this.info.length) {
+		if (name === "colorInfo") {
+			if (!this.colorInfo) {
 				return;
 			}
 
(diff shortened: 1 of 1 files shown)


DmitrySharabin:
I'd like to discuss what is more helpful (for our users) to expose as `colorInfo` keys: the coords labels or channels themselves?

### Option 1

```js
{
    Lightness: 0.9158,
    Chroma: 0.05112,
    Hue: 349.9
}
```

Option 2

```js
{
    "oklch.l": 0.9158,
    "oklch.c": 0.05112,
    "oklch.h": 349.9
}
```

LeaVerou:
I would vote for the keys, the labels might even be localized or have weird characters.

DmitrySharabin:
> I would vote for the keys, the labels might even be localized or have weird characters.

Thank you. I'm leaning toward that, too. Let me update the code, then.

DmitrySharabin:
Done. Here is an updated gist of the proposed changes:

image

src/color-swatch/color-swatch.js
```diff
@@ -189,6 +180,35 @@ const Self = class ColorSwatch extends ColorElement {
 			},
 			dependencies: ["color"],
 		},
+		infoLabels: {
```
LeaVerou, line 183: Why do we need a separate infoLabels prop?

DmitrySharabin: Initially, I was thinking of optimizing the code so that we don't have to use `Object.entries` every time we need to get a channel and its label. It turned out that this was not a good idea: actually, we needed to do this in one place, so there was no need to add an extra prop. I removed it. Thank you for the question; it helped me see the flaw.

LeaVerou (approved):
LGTM

PR #119, opened 2024-10-16
[color-swatch] Add support for color difference (Delta E) and contrast

DmitrySharabin:
(in terms of design, related to #96; and is also its evolution)

The main features are:

- Add support for the `vs` color to calculate deltas and contrast against
- The `info` attribute now allows specific algorithms to be used to calculate DeltaE and contrast. Some usage examples: `info="deltaE2000, WCAG21 contrast"`, `info="Contrast: Michelson contrast"`, `info="DeltaE: deltaEITP, DeltaPhi: DeltaPhi contrast"` (when specifying the contrast method, the word `contrast` is mandatory)
- Add some new props (`infoCoords`, `infoOther`, etc) to get the advantage of caching (provided by `NudeElement`) and make code more DRY
- The code is more forgiving when wrong coords and/or algorithms for deltas and contrast are specified
- Not a breaking change


LeaVerou:
Why do we have two different syntaxes to parameterize what algorithm is used? For ΔE we use the function name (`deltaE2000`) but for contrast we use space separated identifiers. We should pick one syntax and go with it.

Also, thinking of how extends to the color scale case, do we want to be duplicating colors all over the place? Or do we want some way to "link" to another `<color-swatch>` somewhere else? But then how does this work in the `<color-scale>` case where there is no (light DOM) `<color-swatch>` to link to?

DmitrySharabin:
> Why do we have two different syntaxes to parameterize what algorithm is used? For ΔE we use the function name (`deltaE2000`) but for contrast we use space separated identifiers. We should pick one syntax and go with it.

I just thought of another option. What if we adopt the way we used to specify color space and coordinate in one go and do the same thing with ΔE and contrast? For example, `info="deltaE.2000, contrast.WCAG21"`. It's aligned with the rest of the `info` value (if specified) and makes it look more “balanced,” if I can say so.

> Also, thinking of how extends to the color scale case, do we want to be duplicating colors all over the place? Or do we want some way to "link" to another `<color-swatch>` somewhere else? But then how does this work in the `<color-scale>` case where there is no (light DOM) `<color-swatch>` to link to?

Interesting. I need to think about it.

LeaVerou:
> > Why do we have two different syntaxes to parameterize what algorithm is used? For ΔE we use the function name (`deltaE2000`) but for contrast we use space separated identifiers. We should pick one syntax and go with it.
> 
> I just thought of another option. What if we adopt the way we used to specify color space and coordinate in one go and do the same thing with ΔE and contrast? For example, `info="deltaE.2000, contrast.WCAG21"`. It's aligned with the rest of the `info` value (if specified) and makes it look more “balanced,” if I can say so.

I like it!

DmitrySharabin:
Updated usage examples: `info="deltaE.2000, contrast.WCAG21"`, `info="Contrast: contrast.Michelson"`, `info="DeltaE: deltaE.ITP, DeltaPhi: contrast.DeltaPhi"`

src/color-swatch/README.md
```diff
@@ -131,7 +131,7 @@ future_swatch_container.append(swatch);
 ### The `info` attribute
 
 You can use the `info` attribute to show information about the color.
-Currently, the only type of information supported is color coords (in any color space), but more will be added in the future.
+Currently, the types of information supported are color coords (in any color space), the difference (delta) and contrast between the current color and another one (specified via [the `vs` attribute](./#the-vs-attribute)).
```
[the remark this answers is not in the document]
LeaVerou: That said, we may want to think about how to support actual deltas (e.g. hue difference)

src/color-swatch/README.md
```diff
@@ -155,6 +155,25 @@ The `info` attribute plays quite nicely with the `--details-style: compact` styl
 <color-swatch size="large" info="oklch.l, oklch.c, oklch.h" style="--details-style: compact">oklch(70% 0.25 138)</color-swatch>
 ```
 
+### The `vs` attribute
+
+You can calculate the difference (delta) and contrast between the current color and another one.
```
LeaVerou, line 160:
```suggestion
You can calculate the difference (deltaE) and contrast between the current color and another one.
```

src/color-swatch/README.md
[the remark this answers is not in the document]
DmitrySharabin: We already do it in the PR (see the `colorDeltas` prop). For hues, we constrain angles using the shorter arc.

LeaVerou (commented):
Could you add comments above the new props about what kind of info they include? I haven't looked at this in a while and I'm a bit lost

DmitrySharabin:
> Could you add comments above the new props about what kind of info they include? I haven't looked at this in a while and I'm a bit lost

Done. Please let me know if I need to add some more.

src/color-swatch/README.md
```diff
@@ -338,6 +358,8 @@ These properties are read-only.
 | `--transparency-cell-size` | `<length>` | The size of the cells of the transparency gradient. |
 | `--transparcency-background` | `<color>` | The background color of the transparency gradient. |
 | `--transparency-darkness` | `<percentage>` | The opacity of the black color used for dark parts of the transparency gradient. |
+| `--positive-delta-color` | `<color>` | The color used for the positive color difference in color coords. |
+| `--negative-delta-color` | `<color>` | The color used for the negative color difference in color coords. |
```
LeaVerou, line 362: I don't understand what these mean.

src/color-swatch/color-swatch.css
```diff
@@ -7,6 +7,8 @@
 		0 0 / calc(2 * var(--_transparency-cell-size)) calc(2 * var(--_transparency-cell-size))
 		content-box border-box var(--_transparency-background)
 	);
+	--_positive-delta-color: var(--positive-delta-color, hsl(120, 80%, 25%));
+	--_negative-delta-color: var(--negative-delta-color, hsl(0, 85%, 40%));
```
LeaVerou, line 11: Nit: you don't need commas.

src/color-swatch/color-swatch.js
```diff
@@ -207,31 +225,189 @@ const Self = class ColorSwatch extends ColorElement {
...
+					catch (e) {
+						console.error(e);
+					}
+				}
+
+				return ret;
+			},
+		},
+		/**
+		 * Specified deltaE and contrast
```
LeaVerou, line 291: should this be "or contrast"?

PR #131, opened 2024-10-17
[color-chart] Don't show the explicit X-coord in the label, fixes #127

### LINKED ISSUE #127: [color-chart] If we include an explicit `x` via a slash it shouldn't be shown in the label
For now, we have
image
which is wrong.

DmitrySharabin:
image

### DIFF
--- a/src/color-chart/color-chart.js
+++ b/src/color-chart/color-chart.js
@@ -109,6 +109,7 @@ const Self = class ColorChart extends ColorElement {
 		let ret = {
 			element: colorScale,
 			swatches: new WeakMap(),
+			labels: new WeakMap(),
 			x: {min: Infinity, max: -Infinity, values: new WeakMap() },
 			y: {min: Infinity, max: -Infinity, values: new WeakMap()},
 			colors: colorScale.computedColors?.slice() ?? [],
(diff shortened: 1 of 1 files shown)


src/color-chart/color-chart.js
```diff
@@ -128,7 +129,22 @@ const Self = class ColorChart extends ColorElement {
 			ret.colors[index] = color = color.to(this.space);
 			ret.swatches.set(color, swatch);
 
-			let x = Number(name.match(/-?\d*\.?\d+$/)?.[0] ?? index);
+			let x = name.match(/-?\d*\.?\d+$/)?.[0];
+			if (x !== undefined) {
+				// Transform `Label / X-coord` to `Label`
+				// (there should be at least one space before and after the slash so the number is treated as an X-coord)
+				let regexp = new RegExp(`\\s+\\/\\s+${x}$`);
+				name = name.replace(regexp, "").trim();
```
LeaVerou, line 137: We now have the number, so we can match more efficiently, no need to check for *any* number. We can just slice by the length of the number, trim, and then check if that ends with a slash.

LeaVerou (approved):

src/color-chart/color-chart.js
DmitrySharabin: Makes perfect sense. I fixed it. Thanks!

PR #133, opened 2024-10-18
[color-chart] Correctly plot colors without labels, fixes #103

### LINKED ISSUE #103: [color-chart] Doesn't work properly when no labels are provided
We resolved that it should default to the color index, but it seems to be incorrect.
image

DmitrySharabin:
The main idea behind the fix is that we are not interested in using _any_ number on the default label as the X-coord. Such a number _must_ be specified as a coord: either it's the only number in the color label or detached from the previous text with a space.

image

Live preview: https://deploy-preview-133--color-elements.netlify.app/src/color-chart/

### DIFF
--- a/src/color-chart/README.md
+++ b/src/color-chart/README.md
@@ -10,7 +10,7 @@ Display lists of colors as a scatterplot or line chart.
 
 ### Basic usage
 
-Plotting a single color scale (buggy order, see [#103](https://github.com/color-js/elements/issues/103)):
+Plotting a single color scale:
 
 ```html
 <color-chart y="oklch.l">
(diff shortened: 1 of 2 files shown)


LeaVerou:
That seems quite flimsy, and would break things like `Red 100: <color>, Red 200: <color>, ...`.

DmitrySharabin:
> That seems quite flimsy, and would break things like `Red 100: <color>, Red 200: <color>, ...`.

It shouldn't. [None of the charts](https://deploy-preview-133--color-elements.netlify.app/src/color-chart/) we have now is broken:

image

I might miss something, but if you think this is not the fix we want, I'll continue working on the other one (namely, `getX()`).

LeaVerou:
Oooohhhh because they're followed by a space! 

It still seems like a hack, but if it works well, let's do it. Please leave a comment above it explaining why the additional constraint?

LeaVerou (approved):

Issue #120, opened 2024-10-16
[color-chart] Add support for `ymin` and `ymax` props


DmitrySharabin:
As @LeaVerou mentioned in our conversation, we need a way to control whether the bounds the chart is built within are “tight” (the current behavior), i.e., determined by the data, encompassing the whole range of the coordinate or a custom min and max.

It can be accomplished by supporting two new props: `ymin` and `ymax` (these names seem more readable than possible `miny` and `maxy`).

They should (?) support two keywords: `auto` and `coord` (`range`?)—and numbers as values.

### Usage examples

`<color-chart y="oklch.h">...</color-chart>` — `ymin="auto"`, `ymax="auto"` _(default)_

`<color-chart y="oklch.h" ymin="20">...</color-chart>`

`<color-chart y="oklch.h" ymax="135">...</color-chart>`

`<color-chart y="oklch.h" ymin="20" ymax="135">...</color-chart>`

`<color-chart y="oklch.h" ymin="coord">...</color-chart>` — the minimum value of the Y-axis should be taken from the coord’s (in this case, `oklch.h`) range. `ymin` might be less than the minimum value for all color points in the chart.

**Q:** What type should these props have?

I wonder if there is a way to extend the built-in `Number` class to support two new props (`auto` and `coord`) similarly to `NaN`. Or we can probably declare our own class that works with numbers like the built-in one `Number` class but has two extra properties?

DmitrySharabin:
I'm leaning towards defining these props something like this:

```js
ymin: {
	convert (value) {
		// Is a supported keyword?
		if (["auto", "coord"].includes(value)) {
			if (value === "coord") {
				let range = this.yResolved.range ?? this.yResolved.refRange;
				return range?.[0];
			}

			return value;
		}

		return Number(value);
	},
},
```

Is it illegal?

DmitrySharabin:
I have a prototype to discuss (ready to submit a draft PR so we can iterate). Here is it in action:


https://github.com/user-attachments/assets/4ca060f7-a041-4611-a22f-71843a4dde9e

And another one question:

![image](https://github.com/user-attachments/assets/ab79e2e1-2c0a-4103-a7c5-587af4705d38)

LeaVerou:
Awesome, thanks!

I wouldn't expect them to be _disconnected_. I'd expect them to have lines that go out of range, in the direction of the data points. Essentially drawing the whole thing, but applying `overflow: hidden` to the chart.

[...]

PR #134, opened 2024-10-22
[color-chart] First stab at adding `ymin` and `ymax`

DmitrySharabin:
There's [an issue with tooltips](https://github.com/color-js/elements/issues/120#issuecomment-2428769034) that I don't know how to address yet—I'm still experimenting.

### DIFF
--- a/src/color-chart/color-chart.js
+++ b/src/color-chart/color-chart.js
@@ -114,7 +114,10 @@ const Self = class ColorChart extends ColorElement {
 		};
 
 		colorScale.style.setProperty("--color-count", ret.colors.length);
+
 		let yAll = ret.colors.map(({color}) => color.to(this.space).get(this.y));
+		let yMin = this.ymin === "auto" ? -Infinity : this.ymin;
+		let yMax = this.ymax === "auto" ? Infinity : this.ymax;
 
 		if (this.yResolved.type === "angle") {
 			// First, normalize
@@ -156,10 +159,14 @@ const Self = class ColorChart extends ColorElement {
 			ret.x.values.set(color, x);
 			ret.y.values.set(color, y);
 
-			ret.x.min = Math.min(ret.x.min, x);
-			ret.x.max = Math.max(ret.x.max, x);
-			ret.y.min = Math.min(ret.y.min, y);
-			ret.y.max = Math.max(ret.y.max, y);
+			let outOfRange = (isFinite(yMin) && y < yMin) || (isFinite(yMax) && y > yMax);
+			if (!outOfRange) {
+				// Only swatches that are in range participate in the min/max calculation
+				ret.x.min = Math.min(ret.x.min, x);
+				ret.x.max = Math.max(ret.x.max, x);
+				ret.y.min = Math.min(ret.y.min, y);
+				ret.y.max = Math.max(ret.y.max, y);
+			}
 
 			swatch.style.setProperty("--x", x);
 			swatch.style.setProperty("--y", y);
@@ -198,7 +205,7 @@ const Self = class ColorChart extends ColorElement {
 	propChangedCallback (evt) {
 		let {name, prop, detail: change} = evt;
 
-		if (name === "resolvedX" || name === "yResolved") {
+		if (name === "yResolved" || name === "ymin" || name === "ymax") {
 			// Re-render swatches
 			this.render(evt);
 		}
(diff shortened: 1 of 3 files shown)


src/color-chart/color-chart.js
```diff
@@ -232,6 +239,46 @@ const Self = class ColorChart extends ColorElement {
 			// rawProp: "coord",
 		},
 
+		yRange: {
+			get () {
+				return this.yResolved.range ?? this.yResolved.refRange ?? [0, 100];
+			},
+		},
+
+		ymin: {
```
LeaVerou, line 248:
```suggestion
		yΜin: {
```

src/color-chart/color-chart.js
```diff
@@ -232,6 +239,46 @@ const Self = class ColorChart extends ColorElement {
...
+				try {
+					return Number(value);
+				}
+				catch (e) {
+					return "auto";
+				}
+			},
+		},
+
+		ymax: {
```
LeaVerou, line 265:
```suggestion
		yΜax: {
```

LeaVerou (approved):
Just a nit about the prop names, everything else LGTM. Great work!!

DmitrySharabin:
OK, I followed @LeaVerou's [suggestion to use the Popover API](https://github.com/color-js/elements/issues/120#issuecomment-2438510603) to overcome the `overflow: hidden` limitations. It also uses the anchor positioning under the hood.

@LeaVerou, could you please take a look at the changes I made? I'll try to point them out for you in the comments.

src/color-chart/color-chart.js
```diff
@@ -156,15 +159,29 @@ const Self = class ColorChart extends ColorElement {
...
 
+			if (HTMLElement.prototype.hasOwnProperty("popover") && !swatch._el.wrapper.hasAttribute("popover")) {
+				// The Popover API is supported
+				swatch._el.wrapper.setAttribute("popover", "");
+
+				// We need these for the popover to be correctly activated and positioned,
+				// otherwise, it won't be on the top layer
+				swatch.addEventListener("pointerenter", evt => evt.target._el.wrapper.showPopover());
+				swatch.addEventListener("pointerleave", evt => evt.target._el.wrapper.hidePopover());
+			}
```
DmitrySharabin, line 183: Here is the first change

src/color-swatch/color-swatch.css
```diff
@@ -14,6 +14,8 @@
 	width: fit-content;
 	margin: .3em;
 	border-radius: .2rem;
+
+	anchor-name: --color-swatch-anchor;
```
DmitrySharabin, line 18: The second one

src/color-swatch/color-swatch.css
```diff
@@ -129,10 +131,28 @@ slot {
...
+			@supports (position-anchor: --color-swatch-anchor) {
+				/* Bring the popover back on the screen */
+				inset: unset;
+
+				position-anchor: --color-swatch-anchor;
+				position-area: y-start center;
+				margin-bottom: calc(var(--_pointer-height) * 1.6);
+			}
+		}
+
```
DmitrySharabin, line 155: And the last one

LeaVerou:
What is the experience if anchor positioning is not supported?

DmitrySharabin:
> What is the experience if anchor positioning is not supported?

After experimenting with it (and almost pulling all my hair out of all the browser crashes), I decided to drop the support of anchor positioning in favor of the predictable, manual positioning of the popover.

Could you please give this PR another shot when you have some spare cycles?

src/color-chart/color-chart.js
```diff
@@ -232,6 +258,52 @@ const Self = class ColorChart extends ColorElement {
 			// rawProp: "coord",
 		},
 
+		yRange: {
+			get () {
+				return this.yResolved.range ?? this.yResolved.refRange ?? [0, 100];
```
LeaVerou, line 263: Shouldn't this return the actual range if ymin/ymax are used?

DmitrySharabin: We use it to calculate `yMin` and `yMax` when set to `coord`. So, it's `yMin`/`yMax`, depending on range, not vice versa.

LeaVerou: Then I'm not sure we need this prop? It seems like it would be confusing for users, and it's not like it caches some particularly complex calculation.

DmitrySharabin:

Is there anything else we should fix before merging this PR?

LeaVerou:
I wonder if we need to separate the specified `yMin`/`yMax` from the resolved `yMin`/`yMax`. Specifying `ymax="120"` is _very_ different than 120 just being the max of the range. Right now, there is no way to tell because `element.yMax` would just return 120. Any precedent from native HTML elements you can think of?

DmitrySharabin:
> I wonder if we need to separate the specified `yMin`/`yMax` from the resolved `yMin`/`yMax`. Specifying `ymax="120"` is _very_ different than 120 just being the max of the range. Right now, there is no way to tell because `element.yMax` would just return 120. Any precedent from native HTML elements you can think of?

I think `<input type=number />` is one of those elements: We can specify its value with the `value` attribute` (as a string) and, depending on whether it can be transformed to a number or not, the resolved value (which can be accessed via `valueAsNumber`) will be either a number or `NaN`.

I agree that separating the specified `yMin`/`yMax` from the resolved `yMin`/`yMax` makes sense. How about naming the resolved ones `yMinAsNumber`/`yMaxAsNumber` to be aligned with the platform?

LeaVerou (approved):

PR #141, opened 2024-10-23
[color-scale] Don't work with the colors prop raw value

DmitrySharabin:


DmitrySharabin:
As we discussed with @LeaVerou, we should fix the root of the problem (not its symptoms) in `NudeElement`. So, I'm going to close it for now.

PR #142, opened 2024-10-24
[color-scale][color-chart] Export the `gamut` part and hide it in charts, fixes #130

### LINKED ISSUE #130: [color-charts] Gamut badges should be hidden
image

The underlying `<color-swatch>`es display gamut badges if the gamut differs from `sRGB`. In color charts, there is no need for them.

DmitrySharabin:


LeaVerou:
Not sure if we should hide it or just make it way smaller.

DmitrySharabin:
> Not sure if we should hide it or just make it way smaller.

I'm not sure if it's a valid option, but might we combine it with the rest of the information we have about the color (label, coords, etc.)?

LeaVerou:
That would be much harder to implement. Can you try making it smaller (via `font-size`) and see how it looks?

DmitrySharabin:
> That would be much harder to implement. Can you try making it smaller (via `font-size`) and see how it looks?

Here is how it looks with `font-size: 50%`:

image

LeaVerou:
I think even smaller than that might work (40%?), and let's make sure to expose it as a part on `color-chart` so that people can easily hide it.

DmitrySharabin:
> I think even smaller than that might work (40%?)

I think it does indeed:

image


> let's make sure to expose it as a part on `color-chart` so that people can easily hide it.

If I get it right, we have it in this PR.

LeaVerou (approved):
LGTM but let's fix the text color on that gamut badge!

Issue #123, opened 2024-10-16
[New WC] Color space dropdown


DmitrySharabin:
This component automatically gets populated with all color spaces (or a subset, e.g., polar or perceptually uniform, or even a custom subset).

`<color-chart>`, `<color-scale>`, and `<color-picker>` can benefit from it by integrating it and providing an easy way to choose (and dynamically change) the color space. Right now, it requires a lot of plumbing.

LeaVerou:
Name brainstorming:
- `<color-space>` (too generic? is it possible we may want to do something else with this name?)
- `<color-space-picker>` (too long?)

Wrt specifying the spaces:
1. I think a good MVP would be all color spaces 
2. Going slightly beyond MVP, specific ids would be a good escape hatch for making custom things possible
3. Going beyond that, some way to filter by coordinate name and/or type? `:has(h)`? 
	- And that paves the way for other filters too down the line, e.g. `:from(lab)` for all spaces that descend from `lab`
5. Partial ids? E.g. `h*, *h` basically encompasses all polar spaces, `*lab` encompasses both `lab` and `oklab`

The component should do double duty and function as a label for the color space, with clear affordances for clicking to change it. An MVP would be to just use a `<select>`.

DmitrySharabin:
> Name brainstorming:
> 
> * `<color-space>` (too generic? is it possible we may want to do something else with this name?)
> * `<color-space-picker>` (too long?)

By analogy with `<gamut-badge>`, how about `<space-picker>`?

I already have a working MVP, so if we have a name, it would help.

DmitrySharabin:
> Wrt specifying the spaces:
> 
> 1. I think a good MVP would be all color spaces
> 2. Going slightly beyond MVP, specific ids would be a good escape hatch for making custom things possible


I came up with this MVP (I use `SpacePicker` here for brevity; we change it easily for whatever we want):

```js
import ColorElement from "../common/color-element.js";

const Self = class SpacePicker extends ColorElement {
	static tagName = "space-picker";
	static url = import.meta.url;
	static shadowStyle = true;
	static shadowTemplate = `<select id="picker" part="picker"></select>`;

	constructor () {
		super();

		this._el = {};
		this._el.picker = this.shadowRoot.querySelector("#picker");
	}

	connectedCallback () {
		super.connectedCallback?.();
		this._el.picker.addEventListener("change", this);
	}

	disconnectedCallback () {
		this._el.picker.removeEventListener("change", this);
	}

[...]

PR #143, opened 2024-10-24
[space-picker] First iteration

DmitrySharabin:
Partially addresses #123.

### Usage examples

```html
<space-picker></space-picker>
```

```html
<space-picker value="oklab"></space-picker>
```

```html
<space-picker spaces="oklch, p3, srgb"></space-picker>
```

```html
<space-picker spaces="oklch, p3, srgb" value="p3"></space-picker>
```

```html
<space-picker onspacechange="this.nextElementSibling.textContent = this.value"></space-picker>
<output></output>
```

### DIFF
--- a/src/space-picker/space-picker.js
+++ b/src/space-picker/space-picker.js
@@ -1,0 +1,121 @@
+import ColorElement from "../common/color-element.js";
+
+const Self = class SpacePicker extends ColorElement {
+	static tagName = "space-picker";
+	static url = import.meta.url;
+	static shadowStyle = true;
+	static shadowTemplate = `<select id="picker" part="picker"></select>`;
+
+	constructor () {
+		super();
+
+		this._el = {};
+		this._el.picker = this.shadowRoot.querySelector("#picker");
+	}
+
+	connectedCallback () {
+		super.connectedCallback?.();
+		this._el.picker.addEventListener("change", this);
+	}
+
+	disconnectedCallback () {
+		super.disconnectedCallback?.();
+		this._el.picker.removeEventListener("change", this);
+	}
+
+	handleEvent (event) {
+		if (event.type === "change" && event.target === this._el.picker) {
+			this.value = this._el.picker.value;
+		}
+
+		this.dispatchEvent(new event.constructor(event.type, {...event}));
+	}
+
+	propChangedCallback ({name, prop, detail: change}) {
+		if (name === "spaces") {
+			this._el.picker.innerHTML = Object.entries(this.spaces)
+				.map(([id, space]) => `<option value="${id}">${space.name}</option>`)
+				.join("\n");
+		}
+
+		if (name === "value" || name === "spaces") {
+			if (this.value?.id !== this._el.picker.value) {
+				if (this.value.id in this.spaces) {
+					this._el.picker.value = this.value.id;
+				}
+				else {
+					let currentValue = this.spaces[this._el.picker.value];
+					console.warn(`No color space with id = “${ this.value.id }” was found among the specified ones. Using the current one (${ currentValue.id }) instead.`);
+					this.value = currentValue;
+				}
+			}
+		}
+	}
+
+	static props = {
+		value: {
+			default () {
+				return Self.Color.Space.get(this._el.picker.value);
+			},
+			parse (value) {
+				if (value instanceof Self.Color.Space || value === null || value === undefined) {
+					return value;
+				}
+
+				value += "";
+
+				return Self.Color.Space.get(value);
+			},
+			stringify (value) {
+				return value?.id;
+			},
+		},
+
+		spaces: {
+			type: {
+				is: Object,
+				get values () {
+					return Self.Color.Space;
+				},
+				defaultValue: (id, index) => {
+					try {
+						return Self.Color.Space.get(id);
+					}
+					catch (e) {
+						console.error(e);
+					}
+				},
+			},
+			default: () => Self.Color.spaces,
+			convert (value) {
+				// Drop non-existing spaces
+				return Object.fromEntries(Object.entries(value).filter(([id, space]) => space));
+			},
+			stringify (value) {
+				return Object.entries(value).map(([id, space]) => id).join(", ");
+			},
+		},
+	};
+
+	static events = {
+		change: {
+			from () {
+				return this._el.picker;
+			},
+		},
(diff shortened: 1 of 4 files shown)


src/space-picker/space-picker.js
```diff
@@ -0,0 +1,127 @@
...
+					}
+					catch (e) {
+						console.error(e);
+					}
+				},
+			},
+			default: () => Self.Color.spaces,
+			convert (value) {
+				// Drop non-existing spaces
+				let spaces = Object.fromEntries(Object.entries(value).filter(([id, space]) => space));
```
LeaVerou, line 92: I don't think we should drop them, I think we should just not list them with a name. Just the id.

LeaVerou: I.e. the name should default to the id.

DmitrySharabin: Makes sense. Will fix it. Thanks!

src/space-picker/space-picker.js
```diff
@@ -0,0 +1,127 @@
...
+			},
+		},
+		spacechange: {
+			propchange: "value",
+		},
+	};
+
+	static formAssociated = {
+		like: el => el._el.picker,
+		role: "combobox",
```
LeaVerou, line 119: Doesn't this default to the element we specified in `like`? If not, it probably should.

src/space-picker/space-picker.js
```diff
@@ -0,0 +1,127 @@
...
+		valueProp: "value",
+		changeEvent: "spacechange",
```
LeaVerou, line 121: Shouldn't this use `change`?

src/space-picker/space-picker.js
```diff
@@ -0,0 +1,127 @@
...
+				return spaces;
+			},
+			stringify (value) {
+				return Object.entries(value).map(([id, space]) => id).join(", ");
+			},
+		},
+	};
+
+	static events = {
+		change: {
```
LeaVerou, line 107: We should also retarget the `input` event.

src/space-picker/space-picker.js
```diff
@@ -0,0 +1,127 @@
...
+				return Self.Color.Space.get(this._el.picker.value);
+			},
+			parse (value) {
+				if (value instanceof Self.Color.Space || value === null || value === undefined) {
+					return value;
+				}
+
+				value += "";
+
+				return Self.Color.Space.get(value);
```
LeaVerou, line 67: This is again the same issue with the string/object props. We need a prop that contains your actual input as a string (`value`) and one that contains the color space object (likely `space` or `selectedSpace`).

src/space-picker/space-picker.js
```diff
@@ -0,0 +1,127 @@
...
+		super.disconnectedCallback?.();
+		this._el.picker.removeEventListener("change", this);
+	}
+
+	handleEvent (event) {
+		if (event.type === "change" && event.target === this._el.picker) {
+			this.value = this._el.picker.value;
+		}
+
+		this.dispatchEvent(new event.constructor(event.type, {...event}));
```
LeaVerou, line 31: Doesn't defining the change event below take care of this?

DmitrySharabin:
Thank you so much for your review! Every comment is on point, and I'll try to address it ASAP.

src/space-picker/space-picker.js
DmitrySharabin: [In our code, we try to default to `computedRole`](https://github.com/nudeui/element/blob/3c3297eed2506200998b43e29eaa0512c1f0bdcc/src/formAssociated.js/defineFormAssociated.js#L35). Unfortunately, it still isn't supported. It's probably under the flag in some browsers (not sure).

image

DmitrySharabin:
@LeaVerou, I addressed all your feedback (thank you for it). Here are the changes I made:

- Use `value` for actual input (as a string) and `selectedSpace` for the corresponding space object
- Support custom grouping of color spaces via the `groups()` prop
- Don't drop non-existent spaces; replace them with the `{ id, name: id }` objects
- Expose the `valuechange` and `spacechange` events
- Retarget the `input` event
- In `formAssociated`, use `change` as the `changeEvent`

Here is an example of using a custom grouping (from your color palettes research):

```html
<space-picker id="space_picker" spaces="oklch, p3, srgb" value = "p3"></space-picker>

<script type="module">
	space_picker.groups = (spaces) => {
		let ret = {polar: {}, rectangular: {}};
		for (let id in spaces) {
			let space = spaces[id];

			if (space.id !== id) {
				continue;
			}

			let isPolar = space.coords.h?.type === "angle";
			ret[isPolar ? "polar" : "rectangular"][id] = space;
		}
		return ret;
	};
</script>
```

image


Could you please take a look at this PR one more time?

src/space-picker/space-picker.js
LeaVerou: We should just use `getComputedAccessibleNode()` instead. It's not like we don't have access to this information at all.

LeaVerou: (in Nude Element, in case that wasn't clear)

DmitrySharabin:
Thank you so much for your feedback! I tried to address all the issues you pointed out. Namely:

* Add the `groupBy()` prop that expects a space and returns the name of a group the space belongs to

Providing grouping looks like so now:

```html
<space-picker id="space_picker" spaces="oklch, p3, srgb" value = "p3"></space-picker>

<script type="module">
	space_picker.groupBy = (space) => {
		let isPolar = space.coords.h?.type === "angle";
		return isPolar? "Polar" : "Rectangular";
	};
</script>
```

image


* Related to the previous one: `groups` is now a computed prop returning an object with all the groups
* Use different UI for cases with grouping and without

I hope I got everything right. If not, I'll be happy to fix it.

src/space-picker/space-picker.js
DmitrySharabin: Got it. Thanks. I'll send a PR with the change.

LeaVerou (approved):
Looks good! Please add a README, and add it to the list of upcoming components.

DmitrySharabin:
> Looks good! Please add a README, and add it to the list of upcoming components.

Thank you! Will do.

DmitrySharabin:
Done! It's alive and can be played with: https://elements.colorjs.io/src/space-picker/ 🎉

Issue #124, opened 2024-10-16
[New WC] Coordinate dropdown


DmitrySharabin:
Depends on #123 

This new WC encapsulates the color space dropdown and can be used inside `<color-picker>` and `<color-chart>` to allow users to set (change) the color coord dynamically.

DmitrySharabin:
As @LeaVerou [mentioned](https://github.com/color-js/elements/pull/147#issuecomment-2464912628), the main issue with the current styles is that  it doesn't communicate the hierarchy:
![image](https://github.com/user-attachments/assets/abcd796e-e80f-4c5d-858d-2de37ede5f86)

I thought, what if we use a chevron instead of a line, like in breadcrumbs? Will it help?

<img width="990" alt="image" src="https://github.com/user-attachments/assets/9deb8131-6fbe-4791-9bb2-e680e32feec2">

LeaVerou:
Too many chevrons, but it could work if the dropdown arrow only appears on hover/focus

DmitrySharabin:
Here are the UIs we considered (until now) for this component in one place.

**Current one**

![image](https://github.com/user-attachments/assets/abcd796e-e80f-4c5d-858d-2de37ede5f86)

**A là breadcrumbs**

<img width="628" alt="image" src="https://github.com/user-attachments/assets/08e3c5c9-bc82-4fad-b8e4-767c99ac46e5">

With some tweaks (not so many chevrons from the very beginning)

https://github.com/user-attachments/assets/6d449c5a-0e7d-46a6-85bc-ddb81c66cef4


**Channels as a button group**

<img width="993" alt="image" src="https://github.com/user-attachments/assets/f9464f81-7605-41b6-95fa-b059d6f5942c">

With some tweaks (color space picker in bold, smaller text font for channels)

<img width="992" alt="image" src="https://github.com/user-attachments/assets/8b535589-6c34-42e3-8dd0-1351a47d1ad0">

**Using `fieldset`** (basic — channels is a button group, compact — channels is a `<select>` element)

<img width="457" alt="image" src="https://github.com/user-attachments/assets/39aa37be-4319-42a5-a860-0bfedd8b4a75">


**All the above + A là breadcrumbs + connector**

![Channel Picker UI](https://github.com/user-attachments/assets/675dee46-b5bb-43ff-9c88-fe967672a576)

LeaVerou:
Should we close this? We do have this component now.

PR #147, opened 2024-11-06
[channel-picker] First iteration

DmitrySharabin:
Partially addresses #124.

It's a bare minimum, just for a start.

Docs: https://deploy-preview-147--color-elements.netlify.app/src/channel-picker/

### DIFF
--- a/src/channel-picker/channel-picker.js
+++ b/src/channel-picker/channel-picker.js
@@ -1,0 +1,109 @@
+import ColorElement from "../common/color-element.js";
+import "../space-picker/space-picker.js";
+import * as dom from "../common/dom.js";
+
+const Self = class ChannelPicker extends ColorElement {
+	static tagName = "channel-picker";
+	static url = import.meta.url;
+	static shadowStyle = true;
+	static shadowTemplate = `
+		<space-picker part="space_picker" exportparts="picker:space_select" id="space_picker"></space-picker>
+		<select id="picker" part="picker"></select>`;
+
+	constructor () {
+		super();
+
+		this._el = dom.named(this);
+
+		// We need to start listening for this event as soon as the <space-picker> is created
+		this._el.space_picker.addEventListener("spacechange", this);
+	}
+
+	connectedCallback () {
+		super.connectedCallback?.();
+
+		this._el.picker.addEventListener("change", this);
+	}
+
+	disconnectedCallback () {
+		super.disconnectedCallback?.();
+
+		this._el.space_picker.removeEventListener("spacechange", this);
+		this._el.picker.removeEventListener("change", this);
+	}
+
+	#render () {
+		let space = this._el.space_picker.selectedSpace;
+		let coords = space?.coords;
+
+		if (space && !coords) {
+			console.warn(`Color space "${ space.name }" has no coordinates.`);
+			return;
+		}
+
+		this._el.picker.innerHTML = Object.entries(coords)
+			.map(([id, coord]) => `<option value="${ id }">${ coord.name }</option>`)
+			.join("\n");
+
+		let channel = this.value?.split(".")[1];
+		if (channel && channel in coords) {
+			// Preserve the channel if it exists in the new space
+			this._el.picker.value = channel;
+		}
+	}
+
+	handleEvent (event) {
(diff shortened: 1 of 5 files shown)


src/channel-picker/channel-picker.js
```diff
@@ -0,0 +1,109 @@
...
+				if (channel && this._el.picker.value !== channel) {
+					this._el.picker.value = channel;
+				}
+			}
+		}
+	}
+
+	static props = {
+		value: {
+			default: "oklch.l",
```
LeaVerou, line 86: We should definitely have props that return the resolved ColorSpace object and the coordinate used

src/channel-picker/channel-picker.js
```diff
@@ -0,0 +1,109 @@
...
+		let channel = this.value?.split(".")[1];
+		if (channel && channel in coords) {
+			// Preserve the channel if it exists in the new space
+			this._el.picker.value = channel;
+		}
+	}
+
+	handleEvent (event) {
+		if ([this._el.space_picker, this._el.picker].includes(event.target)) {
+			if (event.type === "spacechange") {
```
LeaVerou, line 57: This only applies to space_picker, so I'd move it outside the if

src/channel-picker/channel-picker.js
```diff
@@ -0,0 +1,109 @@
...
+		this._el.picker.addEventListener("change", this);
+	}
+
+	disconnectedCallback () {
+		super.disconnectedCallback?.();
+
+		this._el.space_picker.removeEventListener("spacechange", this);
+		this._el.picker.removeEventListener("change", this);
+	}
+
```
LeaVerou, line 34:
```suggestion

	get selectedSpace () {
		return this._el.space_picker.selectedSpace;
	}
```

src/channel-picker/channel-picker.js
```diff
@@ -0,0 +1,109 @@
...
+		}
+
+		this._el.picker.innerHTML = Object.entries(coords)
+			.map(([id, coord]) => `<option value="${ id }">${ coord.name }</option>`)
+			.join("\n");
+
+		let channel = this.value?.split(".")[1];
+		if (channel && channel in coords) {
+			// Preserve the channel if it exists in the new space
+			this._el.picker.value = channel;
```
LeaVerou, line 51: Likely post-MVP, but even if it doesn't, it would be nice to still keep it around somewhere so that switching _back_ to that space preserves your selection

LeaVerou (changes requested):
Great job, only minor comments. 

One thing that should absolutely change though is the styling: It should be obvious that these are two separate selects. They should still be readable as a unit, but their styling should have **affordances** that indicate that they are separate dropdowns. Like, the triangle should still be there, for both of them. Then you also don't need the weird dashed line.

src/channel-picker/channel-picker.js
DmitrySharabin: During the development, I was adding and removing them several times. I decided if you tell me that we need it, I'll bring them back. 🙂

src/channel-picker/channel-picker.js
DmitrySharabin: I have nothing against it. Will do.

src/channel-picker/channel-picker.js
DmitrySharabin: Good idea! Thanks.

DmitrySharabin:

I also made the picker more resistant to and tolerant of user errors. I would be grateful for your following review.

LeaVerou (approved):
Code looks good, only remaining comment is styling: The control still tries to look like a single thing, which looks weird. I'd render it as either separate dropdowns that are side by side or a "segmented" control, where there is a line down the middle. And the border color should be way fainter in any case.

DmitrySharabin:
> I'd render it as either separate dropdowns that are side by side or a "segmented" control, where there is a line down the middle. And the border color should be way fainter in any case.

Thank you for your feedback. I ended up with a "segmented" control. We can improve it with the next iterations.

image

LeaVerou:
Yeah. One issue with the current styling is that it doesn't communicate the hierarchy, that Lightness is a coordinate *of* Oklch, they look like peers.

PR #149, opened 2024-11-12
[channel-picker] Preserve previously selected channel

DmitrySharabin:
Preview: https://deploy-preview-149--color-elements.netlify.app/src/channel-picker/


LeaVerou (approved):
LGTM, but I don't think we actually need to be that diligent, even if it didn't check the coord name, it would've been fine

DmitrySharabin:
> LGTM, but I don't think we actually need to be that diligent, even if it didn't check the coord name, it would've been fine

I was guided by [your comment in the previous PR](https://github.com/color-js/elements/pull/147#discussion_r1831446314). 😜

PR #150, opened 2024-11-12
[channel-picker] Improve styles

DmitrySharabin:
I experimented with the “breadcrumbs” style and considered [@LeaVerou's proposal](https://github.com/color-js/elements/issues/124#issuecomment-2468849003). Here is what I came up with:

https://github.com/user-attachments/assets/6d449c5a-0e7d-46a6-85bc-ddb81c66cef4

Live preview is here: https://deploy-preview-150--color-elements.netlify.app/src/channel-picker/

### DIFF
--- a/src/channel-picker/channel-picker.css
+++ b/src/channel-picker/channel-picker.css
@@ -1,4 +1,6 @@
 :host {
+	--_transition-duration: .2s;
+
 	--border-width: 1px;
 	--border-color: rgb(0 0 0 / .2);
 	--border-radius: .2em;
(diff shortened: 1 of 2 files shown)


LeaVerou (changes requested):
This is nice, but I’m not sure the cost-benefit is there. It needs **a lot** of extra code for relatively small benefit, and I suspect makes it harder for others to customize the style of the component. 
Also, what is `no-space-picker`? Are we using a separate element to display the color space? We should just use a select either way — styled appropriately. 

I would suggest looking for alternative styles that don't require as much CSS code. Maybe look at other UIs, how do they communicate this kind of containment?

DmitrySharabin:
We decided to go another way, so I'm gonna close it.

PR #155, opened 2024-11-13
[color-picker] First stab at integrating `<space-picker>`

DmitrySharabin:
Here is what it looks like:

image

Live preview: https://deploy-preview-155--color-elements.netlify.app/src/color-picker/

### DIFF
--- a/src/color-picker/README.md
+++ b/src/color-picker/README.md
@@ -85,7 +85,7 @@ All attributes are reactive:
 
 <script>
 	let radios = dynamic_picker.querySelectorAll("input[name=space]");
-	radios.forEach(radio => radio.addEventListener("change", evt => dynamic_picker.space = evt.target.value));
+	radios.forEach(radio => radio.addEventListener("change", evt => dynamic_picker.spaceId = evt.target.value));
 </script>
 
 <style>
--- a/src/color-picker/README.md
+++ b/src/color-picker/README.md
@@ -109,16 +109,9 @@ All attributes are reactive:
 
 | Attribute | Property | Property type | Default value | Description |
 |-----------|----------|---------------|---------------|-------------|
-| `space` | `space` | `ColorSpace` &#124; `string` | `oklch` | The color space to use for interpolation. |
+| `space` | `spaceRaw` | `string` | `oklch` | The color space to use for interpolation. |
+| – | `space` | `ColorSpace` | `OKLCh` | Color space object corresponding to the `space` attribute. |
 | `color` | `color` | `Color` &#124; `string` | `oklch(50% 50% 180)` | The current color value. |
-
-### Getters
-
-These properties are read-only.
-
-| Property | Type | Description |
-|----------|------|-------------|
-| `spaceResolved` | `ColorSpace` | Color space object corresponding to the `space` property. |
 
 ### Events
 
--- a/src/color-picker/README.md
+++ b/src/color-picker/README.md
@@ -40,25 +40,15 @@ You can use your component instead of the default color swatch:
 or your own form element instead of the default space picker:
 
 ```html
-<color-picker space="oklch" color="oklch(60% 30% 180)">
-	<select slot="color-space" size="3" class="horizontal">
-		<option value="oklch" selected>OKLCh</option>
-		<option value="hwb">HWB</option>
-		<option value="hpluv">HPLuv</option>
+<color-picker space="oklab" color="oklab(60% -0.12 0)">
+	<select slot="color-space" size="4">
+		<optgroup label="Rectangular Spaces">
+			<option value="lab">Lab</option>
+			<option value="oklab" selected>Oklab</option>
+			<option value="prophoto">ProPhoto</option>
+		</optgroup>
 	</select>
 </color-picker>
-
-<style>
-	select.horizontal {
-		writing-mode: tb;
-		field-sizing: content;
-
-		option {
-			writing-mode: horizontal-tb;
-			padding: .5em;
-		}
-	}
-</style>
 ```
 
 ### Events
@@ -72,30 +62,38 @@ As with other components, you can listen to the `colorchange` event:
 </color-picker>
 ```
 
-<!-- ### Dynamic
+### Dynamic
 
 All attributes are reactive:
 
 ```html
-<label>
-	Space:
-	<select id="space_select" size="3"></select>
-</label>
+<color-picker space="oklch" color="oklch(60% 30% 180)" id="dynamic_picker">
+	<fieldset slot="color-space">
+		<legend>Polar Spaces</legend>
 
-<color-picker id="dynamic_picker" space="oklch" color="oklch(60% 30% 180)"></color-picker>
+		<label>
+			<input type="radio" name="space" value="oklch" checked /> OKLCh
+		</label>
+		<label>
+			<input type="radio" name="space" value="hwb" /> HWB
+		</label>
+		<label>
+			<input type="radio" name="space" value="hsl" /> HSL
+		</label>
+	</fieldset>
+</color-picker>
 
-<script type="module">
-	import Color from "https://colorjs.io/dist/color.js";
-
-	space_select.innerHTML = Object.entries(Color.spaces)
-		.map(([id, space]) => `<option value="${id}">${space.name}</option>`)
-		.join('\n');
-
-	space_select.value = "oklch";
-
-	space_select.oninput = () => dynamic_picker.space = space_select.value;
+<script>
+	let radios = dynamic_picker.querySelectorAll("input[name=space]");
+	radios.forEach(radio => radio.addEventListener("change", evt => dynamic_picker.space = evt.target.value));
 </script>
-``` -->
+
+<style>
+	label + label {
+		margin-inline-start: .3em;
+	}
+</style>
+```
 
 ## Reference
 
--- a/src/color-picker/README.md
+++ b/src/color-picker/README.md
@@ -41,7 +41,7 @@ or your own form element instead of the default space picker:
 
 ```html
 <color-picker space="oklch" color="oklch(60% 30% 180)">
-	<select slot="space-picker" size="3" class="horizontal">
+	<select slot="color-space" size="3" class="horizontal">
 		<option value="oklch" selected>OKLCh</option>
 		<option value="hwb">HWB</option>
 		<option value="hpluv">HPLuv</option>
@@ -104,7 +104,7 @@ All attributes are reactive:
 | Name | Description |
 |------|-------------|
 | (default) | The color picker's main content. Goes into the swatch. |
-| `space-picker` | An element used to set the picker color space. |
+| `color-space` | An element used to set the picker color space. |
 | `swatch` | An element used to provide a visual preview of the current color. |
 
 ### Attributes & Properties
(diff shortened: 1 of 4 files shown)


src/color-picker/README.md
```diff
@@ -48,7 +48,7 @@ As with other components, you can listen to the `colorchange` event:
 </color-picker>
 ```
 
-### Dynamic
+<!-- ### Dynamic
```
LeaVerou, line 51: This is still useful! But we may want to rework it a bit so that it's not literally replicating a space picker.

src/color-picker/color-picker.css
```diff
@@ -4,6 +4,13 @@
 	gap: 1em;
 }
 
+#space_picker {
+	flex: 0 1 100%; /* occupy the entire row */
```
LeaVerou, line 8: Why?

src/color-picker/color-picker.css
```diff
@@ -4,6 +4,13 @@
...
+	padding: initial;
```
LeaVerou, line 9: ???

src/color-picker/color-picker.css
```diff
@@ -4,6 +4,13 @@
...
+	border: none;
```
LeaVerou, line 10: I'd instead make it transparent and show on hover

src/color-picker/README.md
DmitrySharabin: Yeah, I just commented out so we could rework it.

src/color-picker/color-picker.js
```diff
@@ -1,14 +1,17 @@
...
 import * as dom from "../common/dom.js";
 
 const Self = class ColorPicker extends ColorElement {
 	static tagName = "color-picker";
 	static url = import.meta.url;
 	static dependencies = new Set(["channel-slider"]);
+	static globalStyle = new URL("color-picker-global.css", import.meta.url);
 	static shadowStyle = true;
 	static shadowTemplate = `
+		<space-picker id="space_picker" part="space-picker" exportparts="picker: space-picker-base"></space-picker>
```
LeaVerou, line 14: Can we put it in a slot, and listen to input events on the slot? Then people can slot in any arbitrary form element and as long as it produces `input` events, it will work.

src/color-picker/color-picker.js
```diff
@@ -27,14 +30,22 @@ const Self = class ColorPicker extends ColorElement {
 		super.connectedCallback?.();
 		this._el.sliders.addEventListener("input", this);
 		this._el.swatch.addEventListener("input", this);
+		this._el.space_picker.addEventListener("spacechange", this);
```
LeaVerou, line 33: Wait, why does it fire spacechange and not input events? Does it do both? What's the difference? I thought spacechange was only for channel-picker, to distinguish what changed (though even there we may want to repurpose [`inputType`](https://developer.mozilla.org/en-US/docs/Web/API/InputEvent/inputType) for that)

src/color-picker/color-picker.js
```diff
@@ -145,6 +158,11 @@ const Self = class ColorPicker extends ColorElement {
 				return [this._el.sliders, this._el.swatch];
 			},
 		},
+		spacechange: {
+			from () {
+				return this._el.space_picker;
```
LeaVerou, line 163: Hm that will make it hard to slot custom elements, so I'd revert back to glue code

LeaVerou (changes requested):

src/color-picker/color-picker.css
DmitrySharabin: Taking into account the following comment, I change it. This was just for alignment purposes.

src/color-picker/color-picker.css
DmitrySharabin: Good idea! Thanks. Done.

src/color-picker/color-picker.js
DmitrySharabin: Done!

src/color-picker/color-picker.css
DmitrySharabin: I didn't want to change the flex layout we already use to something else or add a wrapper around the sliders and color swatch to position the space picker above the rest of the elements of the picker, so I did this. Now, I switched to the grid layout.

src/color-picker/color-picker.js
DmitrySharabin: Done!

src/color-picker/color-picker.js
DmitrySharabin: > Does it do both? What's the difference?

It does. When the `<space-picker>` value is set programmatically, we can't use the `input` event to handle it. In this case, we indeed don't need it, so I switched to listening to `input` events on the slot, as you suggested. 

>(though even there we may want to repurpose [inputType](https://developer.mozilla.org/en-US/docs/Web/API/InputEvent/inputType) for that)

I tried dispatching the `input` event using the [`InputEvent` constructor](https://developer.mozilla.org/en-US/docs/Web/API/InputEvent/InputEvent) and passing the `inputType` option with the value of `spaceChange` (in the `<space-picker>` `input` event handler). Still, it seems it wasn't passed along. IDK why. 🤷‍♂️

src/color-picker/README.md
DmitrySharabin: I still don't know what the best alternative for this is. Moreover, I have already demonstrated something similar in the Slots section. I'd be grateful for any ideas.

DmitrySharabin:
I believe I addressed all the issues you pointed out (thank you for this), and the PR is ready for another review.

There is one issue left with the dynamic props demo: see https://github.com/color-js/elements/pull/155#discussion_r1842284461

src/color-picker/README.md
```diff
@@ -37,6 +37,30 @@ You can use your component instead of the default color swatch:
 </color-picker>
 ```
 
+or the default space picker:
```
LeaVerou, line 40:
```suggestion
or your own form element instead of the default space picker:
```

src/color-picker/README.md
```diff
@@ -37,6 +37,30 @@ You can use your component instead of the default color swatch:
...
+		<option value="oklch" selected>OKLCh</option>
+		<option value="hwb">HWB</option>
+		<option value="hpluv">HPLuv</option>
+	</select>
+</color-picker>
+
+<style>
+	select.horizontal {
+		writing-mode: tb;
+		field-sizing: content;
```
LeaVerou, line 54: The horizontal select trick is nice, but not a good option for a demo, you want anything that is not relevant to the demo itself to be relatable, not to have people scratching their heads

src/color-picker/color-picker.js
```diff
@@ -50,22 +59,39 @@ const Self = class ColorPicker extends ColorElement {
...
+		else if (this._el.space_picker.contains(source) || this._slots.space_picker.assignedElements().includes(source)) {
+			this.space = event.target.value;
+			this.color = this.color.to(this.space);
+		}
 
 		this.dispatchEvent(new event.constructor(event.type, {...event}));
 	}
 
+	get spaceResolved () {
+		return this._el.space_picker.selectedSpace;
```
LeaVerou, line 71: What would this return if we don’t have a space picker? The space picker is interchangeable UI, we shouldn't depend on it for resolving the color space.

LeaVerou (changes requested):
Almost there, see comments!

src/color-picker/color-picker.js
DmitrySharabin: It looks like the default `<space-picker>` is always there (even if we have some slotted content instead of it). Its value is always in sync with the `space` property. So, the resolved color space should always be correct.

image

What am I missing?

src/color-picker/README.md
DmitrySharabin: Got it. I'll rework it. Thanks!

src/color-picker/README.md
```diff
@@ -37,6 +37,30 @@ You can use your component instead of the default color swatch:
 </color-picker>
 ```
 
+or the default space picker:
+
+```html
+<color-picker space="oklch" color="oklch(60% 30% 180)">
+	<select slot="space-picker" size="3" class="horizontal">
```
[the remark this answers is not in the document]
DmitrySharabin: Agreed. I'll fix it

src/color-picker/color-picker.js
LeaVerou: Hmmmm. It still seems like a hack but ok

src/color-picker/README.md
```diff
@@ -80,6 +102,7 @@ All attributes are reactive:
 | Name | Description |
 |------|-------------|
 | (default) | The color picker's main content. Goes into the swatch. |
+| `color-space` | An element used to set the picker color space. |
```
LeaVerou, line 105:
```suggestion
| `color-space` | An element to display (and if writable, also set) the current color space. If not provided, a `<space-picker>` is used. |
```

(with `<space-picker>` being a link)

src/color-picker/README.md
```diff
@@ -89,12 +112,20 @@ All attributes are reactive:
 | `space` | `space` | `ColorSpace` &#124; `string` | `oklch` | The color space to use for interpolation. |
 | `color` | `color` | `Color` &#124; `string` | `oklch(50% 50% 180)` | The current color value. |
 
+### Getters
+
+These properties are read-only.
+
+| Property | Type | Description |
+|----------|------|-------------|
+| `spaceResolved` | `ColorSpace` | Color space object corresponding to the `space` property. |
```
LeaVerou, line 121: Given that in JS, `spaceResolved` is far more useful than `space`, I wonder if we should make `space` the one with the actual color space (and make it read/write), and `spaceRaw` or something the one that reflects. Is there any precedent in HTML where the object prop is the one with the "nicer" name?

src/color-picker/README.md
```diff
@@ -89,12 +112,20 @@ All attributes are reactive:
...
+
 ### Events
 
 | Name | Description |
 |------|-------------|
-| `input` | Fired when the color changes due to user action, either with the sliders or the color swatch's input field. |
-| `change` | Fired when the color changes due to user action, either with the sliders or the color swatch's input field. |
+| `input` | Fired when the color changes due to user action, either with the space picker, sliders or the color swatch's input field. |
```
LeaVerou, line 127:
```suggestion
| `input` | Fired when the color changes due to user action, such as adjusting the sliders, entering a color in the swatch's text field, or choosing a different color space. |
```

(and same for the one below)

LeaVerou (changes requested):
This PR LGTM based on my earlier feedback, but thinking about it some more, I think we should use `spaceId` instead of `spaceRaw`. That way it always has a meaning, even when the space is set via JS and I think it would simplify a bunch of code.

Other than that, LGTM. Thanks!

DmitrySharabin:
> I think we should use `spaceId` instead of `spaceRaw`. That way it always has a meaning, even when the space is set via JS and I think it would simplify a bunch of code.

Does this mean we don't need to support setting `spaceId` to a space color object and expect it always to be a string?

For example, for `input[type=number]`, we can set its `value` property to a number or to a string (that can be parsed as a number). In both cases, we'll always get a string when getting `value`. However, if we try to set its value to something that can't be parsed as a number, we'll get a warning, and `value` will be equal to `""`. Do we expect to have something like that in our case?

LeaVerou (approved):

Issue #154, opened 2024-11-13
[channel-slider] Move the actual value from the tooltip to the slider label


DmitrySharabin:
**From @LeaVerou:** Move the actual value where the ranges are (e.g. the (0-1)) and put the range next to the coord name (possibly only appearing on hover)? I think the tooltip looks fancy but hides essential info. It doesn't need to look too much like an `input`, only barely enough to communicate the affordance.

<img width="700" alt="image" src="https://github.com/user-attachments/assets/1bddcc97-da28-4fb4-bf77-f05c6acbb923">

PR #160, opened 2024-11-15
[channel-slider] First stab at working on #154

DmitrySharabin:
- The channel range is moved to the right (to the channel label) and is hidden until the slider is hovered or some of its components have focus.
- The current slider value is at the right part of the slider label and is dimmed until the slider is hovered or the value is focused. The tooltip is removed (or do we need to have it too?)

It looks like this:
image

In the context of `<color-picker>`, it looks like this:
image

One thing that bothers me is that the spin buttons are outside of the slider (if we want the value is aligned to the right of the slider):
image

Or is it worth applying `&::-webkit-textfield-decoration-container { flex-flow: row-reverse; }` to place the spin buttons to the left of the number? I'm afraid it might confuse users, though. But I'm not sure.

image

Or simply don't align the value to the right:

image
image

Live preview: https://deploy-preview-160--color-elements.netlify.app/src/channel-slider/

### DIFF
--- a/src/channel-slider/channel-slider.css
+++ b/src/channel-slider/channel-slider.css
@@ -1,12 +1,14 @@
 .color-slider-label {
+	--_transition-duration: var(--transition-duration, 200ms);
+
 	display: grid;
 	grid-template-columns: 1fr auto;
 	gap: .3em;
-	position: relative;
+	align-items: center;
 
 	em {
 		opacity: 60%;
-		transition: opacity .2s;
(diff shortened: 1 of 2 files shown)


LeaVerou (approved):
This is really nice. Some non-blocking comments:

- I think the buttons on the left are fine. I might even add a border when the element is focused, and a fainter one when it’s hovered, to reinforce that this is a text field that is editable. In fact, we could have a faint border there at all times, and then we don't even need to reorder the buttons.

Issue #161, opened 2024-11-17
[color-picker] Changing space throws a slew of errors


sidewayss:
Your own docs page is my example: https://elements.colorjs.io/src/color-picker/#dynamic

- Select "HWB", or any other color space
- Open Dev Tools Console.
- See a whole series of errors thrown, starting with : 
```
channel-slider.js:141 Unknown channel l in space HWB (hwb). Using first channel (Hue) instead.
channel-slider.js:45 TypeError: No "l" coordinate found in HWB. Its coordinates are: h, w, b
```
It seems to change the space successfully on the screen.  I'm not having as much luck with my implementation. When I set the picker's space to a valid string value, nothing updates on the screen.  The color space changes successfully, but the screen doesn't update at all. The sliders are stuck at LCH, and the text beneath the swatch is stuck too.

PR #166, opened 2024-11-19
Channel slider tweaks

DmitrySharabin:
- Don't bombard the user with errors and warnings; addresses #161
We have already warned the user if the channel is not in the color space:
image

**Before**
image

**After**
image

- Tweak the demo so that the channel selector always has a value
**Before**
image

**After**
image


DmitrySharabin:
> I wonder, instead of swallowing legit errors too, could we simply prevent the non-legit errors from being thrown in the first place?

I'll take a look at what we can do with it.

DmitrySharabin:
I addressed the feedback and think this PR is ready for another review. @LeaVerou, could you please give it another shot?

LeaVerou (approved):

Issue #107, opened 2024-06-24
[color-swatch] Decouple color and label concepts


LeaVerou:
The current API of `<color-swatch>` is a bit of a mess: its contents can be either the color, or its label, or even an input, in which case the color is taken from it and the label is editable. If no label is provided, it is the color. What a mess! This causes surprising bugs like the one in #103 .

Some inspiration could be `<option>`: It supports both `value` and `label`. Both default to `textContent`, but you can set either as an attribute. So you can do either `<option label="United States">us</option>` or `<option value="us">United States</option>

Though in a color swatch it’s common to want to display _both_ the color _and_ its label. I think we should display both and users can always hide the relevant parts.

PR #167, opened 2024-11-20
[color-scale] `editable` prop (MVP)

DmitrySharabin:


src/color-scale/color-scale.js
```diff
@@ -21,21 +21,98 @@ const Self = class ColorScale extends ColorElement {
...
 		this._el.swatches.removeEventListener("colorchange", this, {capture: true});
+		this._el.swatches.removeEventListener("click", this, {capture: true});
 	}
 
 	handleEvent (event) {
+		if (event.type === "input" && (!this.editable?.name || !event.target.matches("input[slot=before]"))) {
+			return;
+		}
+
+		let colors = Object.entries(this.colors);
```
LeaVerou, line 41: I’m missing some comments about what this block of code does overall

src/color-scale/color-scale.js
```diff
@@ -21,21 +21,98 @@ const Self = class ColorScale extends ColorElement {
...
+			let addButton = this._el.addButton;
+			if (this.editable?.color) {
+				if (!addButton) {
+					addButton = this._el.addButton = Object.assign(document.createElement("button"), {
+						id: "add-button",
+						part: "add-button",
+						textContent: "Add color",
+					});
+
+					addButton.addEventListener("click", evt => {
```
LeaVerou, line 82: We should abstract the functionality into a method so that authors can trigger it programmatically (whether or not the `editable` class is there)

LeaVerou (changes requested):
See comments. Also I don't see anything about what keywords are used in `editable`

LeaVerou (changes requested):
Some (possibly non-exhaustive) comments:

 
- High level: I would have expected this to involve a lot more changes in `<color-swatch>`. Why is it trying to do everything in `<color-scale>`? A lot of this functionality is useful for individual swatches too. I'd expect mainly the list stuff to be in `<color-scale>`. Also, #107 seems to be causing issues here too (see below).

Delete button:
 - Remove button should be a part
 - I'd use "delete" rather than "remove" consistently
- The delete button should only appear on interaction, otherwise it adds too much visual noise
- The current styling is more "not allowed" rather than "delete"

Docs:
- Don't show the edge cases before the simple syntax. start from simple, then complex. I'd show the boolean `editable` first, and then the values to customize the behavior.

Other:
- Why are colors displayed twice in 
image
 - Good call on making interpolated colors not editable! Post-MVP, we could make the *number* of stops editable.
 - I would style the add button like an outlined swatch with a light gray border and a big + 
 - The only values I see for `editable` are `name` and `color`. How do I specify whether the list itself is editable and how? (add, delete, reorder, possibly with a `list` shorthand for all three)
- I wonder if a better UI for editing the color/name is a pencil icon next to it which makes it editable then. Or some way to choose.
 - Since UI is likely expected to reactively change the value of `edit` to toggle editing, I wonder if we should call it `edit` rather than `editable`. And we should be consistent across the library.

Food for thought:
- It would be good to start thinking how we can support the common need of modifying colors in a scale with a more elaborate UI, such as the one by our `<color-picker>` or even a third-party one, e.g. like 
image

LeaVerou:
I know the change was my idea, but seeing it now, I think `editable` was a better name 🙈

Also, could we add an `editable` attribute on color swatches and use that rather than encoding color-swatch logic here?

DmitrySharabin:
> I know the change was my idea, but seeing it now, I think `editable` was a better name 🙈

No worries—it's an easy change. I see nothing wrong with trying things out and reverting to one of the previous options if the new ones fail.

> Also, could we add an `editable` attribute on color swatches and use that rather than encoding color-swatch logic here?

Sure. It was exactly the next step I was going to take. I just keep the things around so I don't lose progress.

DmitrySharabin:
I think the PR is ready for another iteration. Now, it does even more than before:

- `<color-swatch>` — https://deploy-preview-167--color-elements.netlify.app/src/color-swatch/#the-editable-attribute
- `<color-scale>` — https://deploy-preview-167--color-elements.netlify.app/src/color-scale/#the-editable-attribute

> * Why are colors displayed twice in
> 
> image

The first are the color names, and the second are the colors themselves. Since we are editing colors here, the names are still visible. Anyway, I updated the example to be less confusing.

Still to implement: reordering colors, changing the number of "stops," enabling edit mode with a pencil button next to the color label, and changing colors with a popup. Probably, next iteration?

@LeaVerou, could you please give the PR another shot?

LeaVerou:
I have not yet properly reviewed it, just wanted to share some rendering glitches on Safari iOS:

IMG_1916
IMG_1917
IMG_1918

also, the + button needs and outline so that it looks a bit more visually similar to the swatches

can we make the name editing UI a bit less prominent than the color editing UI? When both are editable, there is currently nothing to guide the eye, and I think when both are editable, the color is edited a lot more

src/color-scale/color-scale.css
```diff
@@ -1,26 +1,85 @@
...
 }
 
 #swatches {
-	display: flex;
-	gap: inherit;
+	display: contents;
+}
+
+[part="add-button"],
+[part="delete-button"] {
```
LeaVerou, line 14: I would suggest using "add button", then people can target them separately too (and you'd be able to simplify this to just `[part~="button"]

src/color-scale/color-scale.js
```diff
@@ -17,31 +20,253 @@ const Self = class ColorScale extends ColorElement {
...
 		this.dispatchEvent(new event.constructor(event.type, {...event}));
 	}
 
 	propChangedCallback ({name, prop, detail: change}) {
-		if (name === "computedColors") {
+		if (name === "computedColors" && !this.editable?.name && !this.editable?.color) {
 			// Re-render swatches
+			// Only if nothing is being edited, otherwise the input would be lost
+			// or, e.g., "red" would be converted to "rgb(100%, 0%, 0%)" right after the typing is done
+			this.render();
```
LeaVerou, line 71: I would suggest instead checking if it's focused (via `document.activeElement` IIRC). It's entirely possible that it's editable but nobody is interacting with it right now, in which case re-rendering would be fine, right?
Though re-rendering seems a very crude reaction to `computedColors` changing, when does this actually happen? Is it when the entire array if overwritten? In that case I guess it makes sense…

src/color-scale/color-scale.js
```diff
@@ -53,17 +278,56 @@ const Self = class ColorScale extends ColorElement {
...
+				else {
+					swatch.editable = false;
+				}
+
+				let deleteButton = swatch.querySelector("[part=delete-button]");
+				if (this.editable?.delete) {
+					if (!deleteButton) {
+						swatch.insertAdjacentHTML("beforeend", `
+							<button part="delete-button" title="Delete color">
+								<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24">
```
LeaVerou, line 319: Where does this icon come from? Also you don't need the SVG namespace for inline SVG.

src/color-scale/color-scale.js
```diff
@@ -149,6 +416,53 @@ const Self = class ColorScale extends ColorElement {
...
+					return {...value, add: true, delete: true, reorder: true};
+				}
+
+				return value;
+			},
+			reflect: {
+				from: true,
+			},
+		},
+		defaultColor: {
```
LeaVerou, line 460: I think this should also accept a single color as a string or `Color` object (via an attribute too). Don't let the complex case get in the way of the simple one :)

src/color-scale/color-scale.js
```diff
@@ -17,31 +20,253 @@ const Self = class ColorScale extends ColorElement {
...
+		}
+
+		return true;
+	}
+
+	addColor (color, name) {
+		let {name: defaultName, color: defaultColor} = this.defaultColor?.() ?? {};
+
+		name ??= defaultName ?? "New color";
+		color ??= defaultColor ?? this.computedColors.at(-1)?.color ?? new Self.Color("#f06");
```
LeaVerou, line 110: Nice try, but I'd use a middle gray as the ultimate fallback :)

LeaVerou (changes requested):
See individual comments. But more big picture, this requires so much extra code, it makes me wonder if it would have been a better idea to have a separate component that is the editable version.

DmitrySharabin:
> See individual comments. But more big picture, this requires so much extra code, it makes me wonder if it would have been a better idea to have a separate component that is the editable version that inherits from this.

Makes perfect sense. Let me move in that direction. How would we call it? `<color-scale-editable>`?

src/color-scale/color-scale.js
DmitrySharabin: > Nice try, but I'd use a middle gray (per OKLab, even if specified as a hex) as the ultimate fallback :)

I was thinking of an easter egg (sort of). 🙃 But yeah, a middle gray would be a good default.

> Also, more performant to specify a color space and array of coords than rely on parsing.

That's a good point. Thanks!

src/color-scale/color-scale.css
DmitrySharabin: Excellent idea! Thanks.

src/color-scale/color-scale.js
DmitrySharabin: I took the icon from https://iconify.design/. I'll add a comment in the code.

> Also you don't need the SVG namespace for inline SVG.

Thanks. I'll fix it.

src/color-scale/color-scale.js
```diff
@@ -17,31 +20,253 @@ const Self = class ColorScale extends ColorElement {
...
+		if (index < 0 || index >= this._el.swatches.children.length) {
+			console.warn(`No color with index "${ index }". The index should be from 0 up to but not including ${ this._el.swatches.children.length }.`);
+			return false;
+		}
+
+		return true;
+	}
+
+	addColor (color, name) {
+		let {name: defaultName, color: defaultColor} = this.defaultColor?.() ?? {};
```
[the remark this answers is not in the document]
DmitrySharabin: Honestly, I don't know why I made `defaultColor` a function. 🤦‍♂️ I'll make it a “simple” JS prop. Thanks for pointing out the issue.

src/color-scale/color-scale.js
DmitrySharabin: Done!

src/color-scale/color-scale.js
DmitrySharabin: > I would suggest instead checking if it's focused (via `document.activeElement` IIRC). It's entirely possible that it's editable but nobody is interacting with it right now, in which case re-rendering would be fine, right?

You are absolutely right. Thank you for the idea. It works brilliantly. 

> Though re-rendering seems a very crude reaction to `computedColors` changing, when does this actually happen? Is it when the entire array if overwritten?

It does happen when `this.colors` is overwritten; you are right.

DmitrySharabin:
I think I addressed all the issues you mentioned (please let me know if I'm not) and am ready to extract the code to two new components: `<color-swatch-input>` and `<color-scale-input>` (or whatever names we'll end up with).

LeaVerou:
Need to review this properly, but a couple things that jumped at me from a very cursory look:
- buttons need `cursor: pointer`. 
- the color-swatch docs still shows editable as slotting an input, rather than using the `editable` attribute. 
- this icon image is a bit too complex. A pencil icon would suffice. And we should probably use FA icons :) 
- This is too visually complex: image I would do a simple icon with the right text color (which we know) and no background until it's hovered. Once hovered, _then_ it could have a single background (e.g. red with a white icon)
- I would rather avoid accent colors that may not work well with the host page if we don't need an accent color. I.e. I'd make this gray image On hover it could preview what the default color is. It would also be nice to show "Add color" in the swatch name area when you hover.

LeaVerou (changes requested):
Still need to review the code, but left some comments about the design/docs.

PR #168, opened 2024-11-20
[channel-slider] Show the spin buttons all the time

DmitrySharabin:
But make them subtly until the user interacts with them.

**Not focused (hovered)**
image

**Focused**
image

Unfortunately, we are a bit tight with options here (except lowering `opacity` or applying filters, as far as I'm aware). But from my perspective, filters are a bit too much in this case.

### DIFF
--- a/src/channel-slider/channel-slider.css
+++ b/src/channel-slider/channel-slider.css
@@ -45,6 +45,7 @@
 			&::-webkit-inner-spin-button {
 				/* Fade out the spin buttons in Chrome and Safari */
 				opacity: .35;
+				filter: contrast(2);
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


LeaVerou (changes requested):
Can we also make the spin buttons have a transparent bg when not interacted with?

DmitrySharabin:
> Can we also make the spin buttons have a transparent bg when not interacted with?

I tried it. No luck.

DmitrySharabin:
OK, the result with `filter: contrast(2)` looks closer to the desired one. Thank you for the hint!

LeaVerou (approved):

PR #170, opened 2024-11-24
[color-chart] First step in `<channel-picker>` integration

DmitrySharabin:
We need to tweak styles here depending on what base style of `<channel-picker>` we choose (we might end up with this: #169). So, it's only the first step. The implemented logic will be the same regardless.

Live preview: https://deploy-preview-170--color-elements.netlify.app/src/color-chart/

### DIFF
--- a/src/color-chart/color-chart.css
+++ b/src/color-chart/color-chart.css
@@ -32,7 +32,6 @@ slot[name="color-channel"]::slotted(*) {
 	justify-self: start;
 
 	margin-block: .5em .7em;
-	/* margin-inline-start: -.5em; */
 	font-size: 130%;
 }
 
--- a/src/color-chart/color-chart.css
+++ b/src/color-chart/color-chart.css
@@ -28,6 +28,7 @@
 
 [part="color-channel"],
 slot[name="color-channel"]::slotted(*) {
+	/* <channel-picker>, or anything acting like one, should occupy the whole row above the chart so as not to mess up with the rest of the layout */
 	grid-column: 1 / -1;
 	justify-self: start;
 
--- a/src/color-chart/color-chart.css
+++ b/src/color-chart/color-chart.css
@@ -1,7 +1,7 @@
 :host {
 	display: grid;
 	grid-template-columns: auto 1fr;
-	grid-template-rows: 1fr auto;
+	grid-template-rows: auto 1fr auto;
 	height: clamp(0em, 20em, 100vh);
 	contain: size;
 	container-name: chart;
@@ -26,9 +26,19 @@
 	}
 }
 
+[part="color-channel"],
+slot[name="color-channel"]::slotted(*) {
+	grid-column: 1 / -1;
+	justify-self: start;
+
+	margin-block: .5em .7em;
+	/* margin-inline-start: -.5em; */
+	font-size: 130%;
+}
+
 #x_axis {
 	grid-column: 2;
-	grid-row: 2;
+	grid-row: 3;
 
 	.ticks {
 		grid-template-columns: repeat(auto-fit, minmax(0, 1fr));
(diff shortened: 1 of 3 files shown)


src/channel-picker/channel-picker.js
```diff
@@ -25,14 +25,14 @@ const Self = class ChannelPicker extends ColorElement {
 	connectedCallback () {
 		super.connectedCallback?.();
 
-		this._el.picker.addEventListener("change", this);
+		this._el.picker.addEventListener("change", this, {capture: true});
```
LeaVerou, line 28: Why are we using `change` here and not `input`?

DmitrySharabin: IDK why I did it. Probably, there was some misunderstanding. I checked the description of both events on MDN and see no reason to listen to `change` instead of `input` (we want to monitor every change of the picker value, not only the committed ones). I'll fix it. Thanks!

src/color-chart/color-chart.js
```diff
@@ -26,28 +30,40 @@ const Self = class ColorChart extends ColorElement {
...
+
+		this._slots = {
+			color_channel: this.shadowRoot.querySelector("slot[name=color-channel]"),
+		};
 	}
 
 	connectedCallback () {
 		super.connectedCallback();
 		this._el.chart.addEventListener("colorschange", this, {capture: true});
+		this._slots.color_channel.addEventListener("change", this, {capture: true});
```
LeaVerou, line 50: This is still change, is that intentional?

DmitrySharabin: No, it's because I'm an idiot. 🤦‍♂️ Sorry about that. Fixed.

src/channel-picker/channel-picker.js
```diff
@@ -25,14 +25,14 @@ const Self = class ChannelPicker extends ColorElement {
 	connectedCallback () {
 		super.connectedCallback?.();
 
-		this._el.picker.addEventListener("change", this);
+		this._el.picker.addEventListener("input", this, {capture: true});
```
LeaVerou, line 28: input usually bubbles. We shouldn't need capture. If we do, we are emitting an input somewhere and forgot to bubble it.

src/color-chart/color-chart.css
```diff
@@ -26,9 +26,18 @@
 	}
 }
 
+[part="color-channel"],
+slot[name="color-channel"]::slotted(*) {
```
LeaVerou, line 30: Can you add a comment? I couldn't tell what this was about and why we do this (granted, I just woke up, but still, it shouldn't need more thought than what I'm currently capable of 😀 )

src/channel-picker/channel-picker.js
DmitrySharabin: You are absolutely right. Fixed. Thank you!

src/color-chart/color-chart.css
DmitrySharabin: :) Makes sense. Done!

src/color-chart/color-chart.js
```diff
@@ -256,7 +276,15 @@ const Self = class ColorChart extends ColorElement {
 
 		yResolved: {
 			get () {
-				return Self.Color.Space.resolveCoord(this.y, "oklch");
+				try {
+					return Self.Color.Space.resolveCoord(this.y, "oklch");
+				}
+				catch {
```
LeaVerou, line 282: Why do we have this new error condition? I don't understand why this is here based on the comment below.

LeaVerou (approved):
LGTM, see question though

src/color-chart/color-chart.js
DmitrySharabin: Thank you for this question! You are absolutely right: We shouldn't handle invalid values of `y` here. I moved the logic to another place. Could you please take a look and see if it's more correct now?

PR #173, opened 2024-11-27
[channel-slider] Add support for the `alpha` channel

DmitrySharabin:
It works in every version of Color.js. It's a spin-off of #63 to simplify the process of adding new features.

Preview: https://deploy-preview-173--color-elements.netlify.app/src/channel-slider/

PR #174, opened 2024-11-27
[color-picker] Add support for `alpha`, closes #57

### LINKED ISSUE #57: [color-picker] Alpha
`<color-picker>` should optionally support alpha (see https://colorjs.io/apps/picker/ )

We could have a boolean `alpha` property for that.
If the picker color includes alpha, but `this.alpha === false`, the generated color would just continue having that alpha.

Though, we can kill two birds with one stone and have a `sliders` or `show` property that controls *all* sliders shown and their order. E.g. `show="l, c, h, alpha"`. 

In the future it could even be extended to include sliders from other color spaces, like sliders="oklch.h, lch.c, hsl.l", though not sure whether there are use cases for that. But the syntax can support it if use cases arise!

DmitrySharabin:
Related to #173 (and should be merged after it). Works with all versions of Color.js

Docs: https://deploy-preview-174--color-elements.netlify.app/src/color-picker/#the-alpha-attribute

### DIFF
--- a/src/color-picker/README.md
+++ b/src/color-picker/README.md
@@ -22,22 +22,10 @@ If no color space or color is provided, the default ones will be used: `oklch` f
 
 ### The `alpha` attribute
 
-Colors with the alpha channel are also supported:
+Colors with the alpha channel are also supported. Add the `alpha` boolean attribute to show the alpha channel:
 
 ```html
-<color-picker space="oklch" color="oklch(60% 30% 180 / .6)"></color-picker>
-```
-
-Add the `alpha` boolean attribute to show the alpha channel permanently:
-
-```html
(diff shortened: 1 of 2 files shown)


DmitrySharabin:
Now, when #173 is merged, we can support alpha and in `<color-picker>`, too.

LeaVerou (changes requested):
What are the use cases for showing the alpha only some of the time? I think this can be quite jarring when people are typing. Unless it is driven by actual user requests I suggest we go with a simple boolean alpha for now.

DmitrySharabin:
> What are the use cases for showing the alpha only some of the time? I think this can be quite jarring when people are typing. Unless it is driven by actual user requests I suggest we go with a simple boolean alpha for now.

I have no objections. Done!

LeaVerou (approved):

PR #187, opened 2024-12-10
[website] Improve the home page

DmitrySharabin:
To represent color components, use cards instead of a plain list.


LeaVerou (commented):
Approved since this is a clear improvement. However, since the components themselves are so colorful, I would suggest making the cards more neutral so the components can shine. 

Nice illustrations! 

Not a huge fan of the icon for upcoming

also having short descriptions would be good

DmitrySharabin:
> However, since the components themselves are so colorful, I would suggest making the cards more neutral so the components can shine.

I was expecting that, so I have plan B. 😅

> Not a huge fan of the icon for upcoming

I'll suggest another version then (probably a few)

LeaVerou:
I think it's better (though I reviewed the prev one from my phone so can't fully compare).

A few comments:
image

Why does this use different styling?

---

I think we want one of the two areas to have a subtly different background. I tried both:
image

And I think the version with the bottom having the subtle gray works better

----

I would increase border-radius on the card, since it's a container, and reduce vertical spacing in the figcaption:

image

(Still haven't looked at the code, this is just feedback about the rendering)

DmitrySharabin:
> (Still haven't looked at the code, this is just feedback about the rendering)

Thank you so much! That was very useful feedback. I'll make the changes you suggested.

> image
> 
> Why does this use different styling?

I was thinking of something that shows that components are not ready yet, but it probably doesn't make much sense—the upcoming components are already in a separate section.

DmitrySharabin:
Done!

LeaVerou (approved):

PR #188, opened 2024-12-17
Entirely hide elements before they are defined

DmitrySharabin:
We should hide not only the element itself but all its descendants in the Light DOM (crucial for `<color-chart>`s).

I went with `opacity: 0` instead of `display: none` because it seems to cause less layout shift.


DmitrySharabin:
It was a bad idea: everything stayed the same, except I ended up with more complex code. 🤦‍♂️

DmitrySharabin:
It was a bad idea: everything stayed the same, except I ended up with more complex code. 🤦‍♂️

PR #189, opened 2024-12-18
[website] Fix horizontal scrolling on pages with tables

DmitrySharabin:
### Changes
- Use different layouts for tables with examples on smaller screens
- Make big tables scrollable and add scrolling hints for better UX

### Remaining issues

On tiny screens, there still might be overflows, causing horizontal scrolling. Like so:
image

I can think of one fix for this (and already tried it): add `overflow-x: auto` to the slot named `demo` of the `html-demo` element. To do so, we should probably make it a part (and `toolbar` and `code` as well). This change is already a part of the element's roadmap, so it's a planned improvement.

Adding `overflow-x: auto`, though, might cause an issue with the details popup:
image

I wonder if we should move [the logic we use in `<color-chart>` to overcome this limitation with `overflow`](https://github.com/color-js/elements/blob/832c1abe895500235740efdb99b80e04fdaade38/src/color-chart/color-chart.js#L191-L208) to the `<color-swatch>` element itself and handle it there? In that case, the issue will be fixed in `<color-swatch>` and `<color-chart>`.

### DIFF
--- a/assets/css/style.css
+++ b/assets/css/style.css
@@ -77,6 +77,25 @@
 	}
 }
 
+.scrollable {
+	--_transparent-canvas: color-mix(in oklab, canvas, oklab(none none none / 0%) 100%);
+	--_hint-color: color-mix(in oklab, canvastext 30%, var(--_transparent-canvas));
+
+	max-inline-size: 100%;
+	overflow-x: auto;
+
+	/**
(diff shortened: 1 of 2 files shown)


LeaVerou:
This ...doesn't seem like the right solution. Ideally, these should not be rendered side by side if there's not enough space to render them side by side…

PR #190, opened 2024-12-19
Attempt to eliminate the flash of undefined custom elements

DmitrySharabin:


src/common/color-element.js
```diff
@@ -39,7 +39,16 @@ const Self = class ColorElement extends NudeElement {
 		// Hide elements before they are defined
 		let style = document.getElementById("color-element-styles")
 		          ?? Object.assign(document.createElement("style"), {id: "color-element-styles"});
-		style.textContent = `:is(${ colorTags.join(", ") }):not(:defined) {display: none}`;
+		style.textContent = `
+			:is(${ colorTags.join(", ") }) {
+				opacity: 0;
+
+				body:not(.loading-global-color-element-styles) &:defined {
```
LeaVerou, line 46: Also, ideally this should be solved in Nude Element (via a mixin): components should define their dependencies via a static property, and then these styles are generated for all dependencies.

src/common/color-element.js
```diff
@@ -80,6 +89,23 @@ const Self = class ColorElement extends NudeElement {
 		}
 
 		customElements.define(this.tagName, this);
+
+		if (this.globalStyle) {
+			// NudeElement imports global styles using `@import url()` inside an `<style>` element with a `data-for` attribute equal to the element's tag name
```
LeaVerou, line 94: Yeah, looks like this was a bit overfit. We should definitely support verbatim styles being included, not just URLs. We could probably even detect the two so the user can just give us an array of strings

DmitrySharabin: That's a very good idea. With something like this do: https://github.com/nudeui/element/pull/43?

PR #195, opened 2025-01-16
[color-swatch][color-scale] Fix layout issues when there are many color swatches

DmitrySharabin:
Also, on small devices.

This fix was part of [another PR](#167), but since we don't want it to be held, I extracted it to this one.

~~This fix allows [color palettes](https://palettes.colorjs.io/palettes/) to look a bit better:~~
image

~~However, not all the issues can be fixed. For example, [Subgrid doesn't play well with CQ](https://codepen.io/dmitrysharabin/pen/oNRZzbb) in all (🫣) browsers! That's why we can't get (at least) something like this by merging this PR:~~
image

I'll try to improve this further, but I just want to ensure we have something to start with.

**UPDATE**

I've come up with another approach I like more: We can apply `text-overflow: ellipsis` to swatch labels on overflow (the full label text can be seen in a tooltip on hover, thanks to the `title` attribute).

image

We also expose it as a `swatch-label` part for alternative styling. For example, one can make the labels scrollable if they like:

```css
color-scale::part(swatch-label) {
	overflow: scroll;
	text-overflow: revert;
	overscroll-behavior: contain;
}
```

image

### DIFF
--- a/src/color-scale/color-scale.css
+++ b/src/color-scale/color-scale.css
@@ -1,26 +1,21 @@
 :host {
+	display: grid;
 	gap: .3em;
+	grid-auto-flow: row;
+	grid-template-rows: auto auto;
+	grid-template-columns: repeat(auto-fit, minmax(0, 1fr));
 }
 
 #swatches {
-	display: flex;
-	gap: inherit;
+	display: contents;
 }
 
 color-swatch {
 	margin: 0;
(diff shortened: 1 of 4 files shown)


LeaVerou (approved):
Might not be perfect, but definitely an improvement!

PR #197, opened 2025-02-28
[color-chart] Fix bug with angle normalization

DmitrySharabin:


src/color-chart/color-chart.js
```diff
@@ -426,9 +426,7 @@ function normalizeAngles (angles) {
 		if (Math.abs(delta) > 180) {
 			let equivalent = [h + 360, h - 360];
 			// Offset hue to minimize difference in the direction that brings it closer to the average
-			let delta = h - averageHue;
-
-			if (Math.abs(equivalent[0] - prevHue) <= Math.abs(equivalent[1] - prevHue)) {
+			if (Math.abs(equivalent[0] - averageHue) <= Math.abs(equivalent[1] - averageHue)) {
```
LeaVerou, line 429: Does averageHue need to be normalized first before we take these deltas?

I wonder if the function should be recursive: Implement a simpler (and faster) version when it's just <= 2 angles, and then use it for the longer version every time you need to subtract hues.

DmitrySharabin: > Does averageHue need to be normalized first before we take these deltas?

`averageHue` is calculated based on normalized hues. Doesn't that mean that it's also normalized?

https://github.com/color-js/elements/blob/7b79fd9eb31b83944b15413323c042793fea5eb4/src/color-chart/color-chart.js#L416-L419


> I wonder if the function should be recursive: Implement a simpler (and faster) version when it's just <= 2 angles, and then use it for the longer version every time you need to subtract hues.

Let me try.

DmitrySharabin: I couldn't think of a simpler recursive variant of the function that would give us the correct result, but I might not understand something. 🤷‍♂️

PR #202, opened 2025-07-28
[color-swatch] Fix popovers with color details in `<color-chart>`s

DmitrySharabin:
Now it works not only in Chrome, but also in Safari and Firefox. [Support for `popover`](https://caniuse.com/?search=popover) is much wider than [support for style queries](https://caniuse.com/css-container-queries-style). For example, Firefox still lacks them.

Open the link in Safari/Firefox for preview: https://deploy-preview-202--color-elements.netlify.app/src/color-chart/

@LeaVerou, could you please take another look at this PR? It turned out that it solves issues not only in Safari.

### DIFF
--- a/src/color-swatch/color-swatch.css
+++ b/src/color-swatch/color-swatch.css
@@ -94,10 +94,6 @@ slot {
 }
 
 [part="details"] {
-	/*
-		Why here and not in the style query below? Because Safari has issues with parsing nested rules inside style queries,
-		we need to duplicate some rules to achieve the desired effect. Moving custom properties here allows us to make the code a bit DRYer.
-	*/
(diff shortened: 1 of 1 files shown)


LeaVerou:
Safari supports style queries, why wasn't it working in Safari?

DmitrySharabin:
> Safari supports style queries, why wasn't it working in Safari?

I still don't have an answer to this question. I can't reproduce the issue. In simple cases, everything works as expected. However, in our case, we have a hierarchy of custom elements: some are within a component's light DOM, while others are in the shadow DOM. And since we rely on many edge features, some may not play well together. I don't have another explanation for now. 🤷‍♂️

PR #206, opened 2025-07-29
Fix charts rendering

DmitrySharabin:
+ Prettier

Although ticks on the custom X axis should be tweaked:
image

And it seems to me that we are facing race conditions sometimes. Reloading the page might end up with:
image

### DIFF
--- a/src/color-chart/color-chart.js
+++ b/src/color-chart/color-chart.js
@@ -30,9 +30,9 @@ const Self = class ColorChart extends ColorElement {
 		super();
 
 		this._el = {
-			slot:   this.shadowRoot.querySelector("slot:not([name])"),
+			slot: this.shadowRoot.querySelector("slot:not([name])"),
 			channel_picker: this.shadowRoot.getElementById("channel_picker"),
-			chart:  this.shadowRoot.getElementById("chart"),
+			chart: this.shadowRoot.getElementById("chart"),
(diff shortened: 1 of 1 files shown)


src/color-chart/color-chart.js
```diff
@@ -81,16 +84,15 @@ const Self = class ColorChart extends ColorElement {
 	 * (Re)render one of the axes
 	 * @param {string} axis "x" or "y"
 	 */
-	renderAxis(axis) {
-
-		let min =  this[`${axis}MinAsNumber`];
-		if (isNaN(min)) {
+	renderAxis (axis) {
+		let min = this[`${axis}MinAsNumber`];
+		if (isNaN(min) || !isFinite(min)) {
```
DmitrySharabin, line 89: This one.

src/color-chart/color-chart.js
```diff
@@ -81,16 +84,15 @@ const Self = class ColorChart extends ColorElement {
...
 			// auto, undefined, etc
 			min = this.bounds[axis].min;
 		}
 
 		let max = this[`${axis}MaxAsNumber`];
-		if (isNaN(max)) {
+		if (isNaN(max) || !isFinite(max)) {
```
DmitrySharabin, line 95: And this one.

LeaVerou (approved):

PR #211, opened 2025-07-31
Follow the provided min/max values, if any

DmitrySharabin:
For now, `<color-chart>` returns a nice-looking axis even if the specified range won't be met. In most cases, it will be a wider one. But users expect to see the exact range they specified. This PR fixes it.

### DIFF
--- a/src/color-chart/color-chart.js
+++ b/src/color-chart/color-chart.js
@@ -101,7 +101,13 @@ const Self = class ColorChart extends ColorElement {
 		}
 
 		if (isFinite(min) && isFinite(max)) {
-			const axisData = getAxis(min, max, 10);
+			const axisData = getAxis({
+				min,
+				max,
+				initialSteps: 10,
+				rawMin: this[`${axis}Min`],
+				rawMax: this[`${axis}Max`],
+			});
 
 			this._el.chart.style.setProperty(`--min-${axis}`, axisData.min);
 			this._el.chart.style.setProperty(`--max-${axis}`, axisData.max);
@@ -478,7 +484,7 @@ const Self = class ColorChart extends ColorElement {
 					let range = this.xResolved?.refRange ?? this.xResolved?.range ?? [0, 100];
 					return range[0];
 				}
-				else if (!this.x || this.xMin === "auto") {
+				else if ((!this.x && isNaN(this.xMin)) || this.xMin === "auto") {
 					return this.bounds.x.min;
 				}
 
@@ -516,7 +522,7 @@ const Self = class ColorChart extends ColorElement {
 					let range = this.xResolved?.refRange ?? this.xResolved?.range ?? [0, 100];
 					return range[1];
 				}
-				else if (!this.x || this.xMax === "auto") {
+				else if ((!this.x && isNaN(this.xMax)) || this.xMax === "auto") {
 					return this.bounds.x.max;
 				}
 
@@ -549,7 +555,7 @@ Self.define();
 
 export default Self;
 
-function getAxis (min, max, initialSteps) {
+function getAxis ({ min, max, initialSteps, rawMin, rawMax }) {
 	let range = max - min;
 	let step = range / initialSteps;
 	let magnitude = Math.floor(Math.log10(step));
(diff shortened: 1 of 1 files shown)


src/color-chart/color-chart.js
```diff
@@ -101,7 +101,13 @@ const Self = class ColorChart extends ColorElement {
 		}
 
 		if (isFinite(min) && isFinite(max)) {
-			const axisData = getAxis(min, max, 10);
+			const axisData = getAxis({
+				min,
+				max,
+				initialSteps: 10,
+				rawMin: this[`${axis}Min`],
```
LeaVerou, line 108: What do we plan to do with the dataset min/max if we have user-provided ones? I don't think we should use them in any way, right? 

If so, I would frame it around `force`, rather than `raw`. 
Either as a two separate settings (`forceMin: true, forceMax: true`) or  one separate setting (`force: {min: true}` with `force: true` for both) or as part of min/max's value (`min: {min, force: true}`

LeaVerou: We would be passing them as `min` and `max` in that case, since we don't need the min and max of the actual dataset (do we?).

DmitrySharabin: Oh, right. Now I got it.

Thank you so much for your excellent feedback. I addressed it and am ready for another iteration.

src/color-chart/color-chart.js
```diff
@@ -101,7 +101,12 @@ const Self = class ColorChart extends ColorElement {
 		}
 
 		if (isFinite(min) && isFinite(max)) {
-			const axisData = getAxis(min, max, 10);
+			const axisData = getAxis({
+				min,
+				max,
+				initialSteps: 10,
+				force: { min: !isNaN(this[`${axis}Min`]), max: !isNaN(this[`${axis}Max`]) }, // follow the provided min/max values, if any
```
LeaVerou, line 108: This is a bit of an obscure condition, but also we check for these earlier (L94 and L100), so we can just set flags there. Or even just a `let force = {min: true, max: true}` and ovverride it in these conditionals, then you don't even need to pass a value here.

src/color-chart/color-chart.js
```diff
@@ -478,7 +486,7 @@ const Self = class ColorChart extends ColorElement {
 					let range = this.xResolved?.refRange ?? this.xResolved?.range ?? [0, 100];
 					return range[0];
 				}
-				else if (!this.x || this.xMin === "auto") {
+				else if ((!this.x && isNaN(this.xMin)) || this.xMin === "auto") {
```
LeaVerou, line 489:
```suggestion
				else if ((!this.x && force.min) || this.xMin === "auto") {
```
Basically, I'm trying to avoid having the logic for when we're forcing min and max spread over in multiple places.

DmitrySharabin: We don't have access to `force` here (it's inside props). Should we make it a prop or an instance property?

src/color-chart/color-chart.js
```diff
@@ -478,7 +507,7 @@ const Self = class ColorChart extends ColorElement {
 					let range = this.xResolved?.refRange ?? this.xResolved?.range ?? [0, 100];
 					return range[0];
 				}
-				else if (!this.x || this.xMin === "auto") {
+				else if ((!this.x && !this.force.x.min) || this.xMin === "auto") {
```
LeaVerou, line 510: While it's not _particularly_ slow, let's not be wasteful; I'd assign `this.force` to a variable rather than invoking the getter every time

src/color-chart/color-chart.js
```diff
@@ -75,6 +75,30 @@ const Self = class ColorChart extends ColorElement {
 
 	bounds = { x: { min: Infinity, max: -Infinity }, y: { min: Infinity, max: -Infinity } };
 
+	// Follow the provided min/max values, if any
+	get force () {
+		const ctx = this;
```
LeaVerou, line 80: What's the point of using `ctx` instead of `this`? It doesn't provide any performance improvement and is only one character shorter, so it doesn't seem like the indirection is worth it.

LeaVerou (approved):
LGTM, see minor nit

src/color-chart/color-chart.js
DmitrySharabin: @LeaVerou, do you mean like so? Does it give us any advantages if we do that in prop getters?

image

Issue #212, opened 2025-08-04
Components should handle complex color values gracefully


LeaVerou:
The set of colors that Color.js can parse and that browsers can parse are overlapping but not identical, and neither is a subset of the other.
Color.js understands color spaces that browsers do not, but also browsers support things like `color-mix()`, CSS variables, relative colors etc.

A big improvement would be to introduce an extra step: If parsing fails, try `CSS.supports('color', color)`. If that is true, use `getComputedStyle()` to resolve color ([demo](https://codepen.io/leaverou/pen/pvjPPVZ?editors=1111)), then proceed as normal

PR #213, opened 2025-08-04
FIrst stab at #212

DmitrySharabin:
[color-swatch] Handle complex color values gracefully

### DIFF
--- a/src/color-swatch/color-swatch.js
+++ b/src/color-swatch/color-swatch.js
@@ -23,10 +23,13 @@ const Self = class ColorSwatch extends ColorElement {
 			<slot name="after"></slot>
 		</div>`;
 
+	static resolvedColors = new Map();
+
 	constructor () {
 		super();
 
 		this._el = {
+			swatch: this.shadowRoot.querySelector("slot[name=swatch]::slotted(*), #swatch"),
 			wrapper: this.shadowRoot.querySelector("#wrapper"),
 			label: this.shadowRoot.querySelector("[part=label]"),
 			colorWrapper: this.shadowRoot.querySelector("[part=color]"),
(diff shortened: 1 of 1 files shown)


src/common/color-element.js
```diff
@@ -137,6 +140,38 @@ const Self = class ColorElement extends NudeElement {
...
+			if (!CSS.supports("color", value)) {
+				// Not supported or invalid value
+				return null;
+			}
+
+			// One of the supported color values; resolve and cache it
+			element.style.backgroundColor = value;
+			let color = getComputedStyle(element).backgroundColor;
+			color = Self.Color.get(color);
+			Self.#resolvedColors.set(value, color);
```
LeaVerou, line 169: One issue with this is that now multiple components can share the same `Color` object if they use the same string, and I'm pretty sure we modify it. So perhaps we should cache the resolved string and create a new instance every time.

src/common/color-element.js
```diff
@@ -137,6 +140,42 @@ const Self = class ColorElement extends NudeElement {
...
+	 * Resolve a color value and cache it.
+	 * @param {string} value Color value to resolve.
+	 * @param {Element} element Element to get computed style from to resolve the color value.
+	 * @returns {Color | null} Resolved color value or null if the value cannot be resolved.
+	 */
+	static resolveColor (value, element) {
+		try {
+			return Self.Color.get(value);
+		}
+		catch {
```
LeaVerou, line 154: You don't need to nest everything in the `catch`. Since the try will return if it succeeds, you can just have an empty catch here and proceed as normal.

src/common/color-element.js
```diff
@@ -137,6 +140,42 @@ const Self = class ColorElement extends NudeElement {
...
+			return Self.Color.get(color);
+		}
+
+		if (!globalThis.CSS?.supports("color", value) || !element) {
+			// Not supported/invalid value, or no element to resolve the color value from
+			return null;
+		}
+
+		// One of the supported color values; resolve and cache it
+		element.style.backgroundColor = value;
```
LeaVerou, line 168: We should store the element's previous `backgroundColor` so we can restore it, rather than overwriting.
Also, we should avoid the DOM I/O if the element *already* has that background (since any DOM operation is slow), so perhaps we can pass in an option for that.

DmitrySharabin: > We should store the element's previous `backgroundColor` so we can restore it, rather than overwriting.

I fixed it. Thanks!

> Also, we should avoid the DOM I/O if the element _already_ has that background (since any DOM operation is slow), so perhaps we can pass in an option for that.

Could you elaborate, please? How do you see it?

LeaVerou: > Could you elaborate, please? How do you see it?

Like, a setting (name TBD) that will tell the function "the element I'm giving you already has the right background, don't bother setting it"

PR #217, opened 2025-09-04
Update import path for NudeElement

domoritz:
Use implicit node modules import so that esm works


LeaVerou:
Closing as no longer relevant. We use specifiers now so you can use whatever colorjs version you want!

PR #220, opened 2025-10-15
Adopt the new `NudeElement` API

DmitrySharabin:
Should be merged *after* https://github.com/nudeui/element/pull/53 is merged and a new version of NudeElement is released.


DmitrySharabin:
This is not the case anymore.

PR #227, opened 2026-05-01
Migrate to the (new) `nude-element` API

DmitrySharabin:


src/channel-picker/README.md
```diff
@@ -64,8 +64,9 @@ All properties are reactive and can be set programmatically:
 		channel_slider.space = space;
 		channel_slider.channel = channel;
 	}
-	
-	channel_picker.onvaluechange = updateSlider;
+
+	channel_picker.addEventListener("valuechange", updateSlider);
+	customElements.whenDefined("channel-picker").then(updateSlider);
```
LeaVerou: Why? This should not be needed. If the original code doesn't work anymore, we shouldn't quietly document workarounds, we should fix it.

src/channel-slider/channel-slider.js
```diff
@@ -256,10 +260,11 @@ const Self = class ChannelSlider extends ColorElement {
 			get () {
 				return this.colorAt(this.value);
 			},
-			dependencies: ["defaultColor", "space", "channel", "value"],
 			set (value) {
 				this.defaultColor = value;
-				this.value = this.defaultValue;
+				// Clear value so it re-engages defaultProp tracking against the new
+				// defaultColor in the next drain. Reading defaultValue here would be stale.
```
LeaVerou: What does this mean? These components are also essentially user-testing nude-element. We should not expect people who just want to define a component to know about these internal implementation details.

src/color-chart/color-chart.js
```diff
@@ -50,11 +50,13 @@ const Self = class ColorChart extends ColorElement {
 		super.connectedCallback();
 		this._el.chart.addEventListener("colorschange", this, { capture: true });
 		this._slots.color_channel.addEventListener("input", this);
+		this._el.channel_picker.addEventListener("valuechange", this);
 	}
 
 	disconnectedCallback () {
 		this._el.chart.removeEventListener("colorschange", this, { capture: true });
 		this._slots.color_channel.removeEventListener("input", this);
+		this._el.channel_picker.removeEventListener("valuechange", this);
```
LeaVerou: Why do we suddenly need these? We did not decide on any such API change, did we? This looks like it's working around a regression.

src/color-chart/color-chart.js
```diff
@@ -377,20 +375,32 @@ const Self = class ColorChart extends ColorElement {
...
 			convert (value) {
-				// Try setting the value to the channel picker. The picker will handle possible erroneous values.
-				this._el.channel_picker.value = value;
-
-				// If the value is not set, that means it's invalid.
-				// In that case, we are falling back to the picker's current value.
-				if (this._el.channel_picker.value !== value) {
-					return this._el.channel_picker.value;
+				try {
+					Self.Color.Space.resolveCoord(value, "oklch");
```
LeaVerou, line 380: Why are we resolving everything against oklch? We want to be able to plot any coordinate, in any color space! The color spaces of X and Y don't need to match, either.

src/color-chart/color-chart.js
```diff
@@ -377,20 +375,32 @@ const Self = class ColorChart extends ColorElement {
...
+					return value;
 				}
-
-				return value;
+				catch (err) {
+					// Try a same-space fallback before giving up entirely.
```
LeaVerou: What is the error condition here? Also, this change looks unrelated to migrating to the new nude-element API?

PR #231, opened 2026-06-03
Add `gamut` support to `<color-slider>` and `<channel-slider>`

LeaVerou:
## Summary

Adds out-of-gamut visualization to `<color-slider>` and `<channel-slider>`. When a `gamut` is set, the parts of the slider's band whose interpolated color falls outside that gamut are overlaid with a customizable gray, so you can see at a glance which values are actually reproducible (e.g. in sRGB).

## `<color-slider>`

- **`gamut`** — a color.js gamut id (`srgb`, `p3`, `rec2020`), or **`auto`** to follow the display's gamut (detected via the `color-gamut` media query, and updated live if the display changes, e.g. the window moves to another monitor). `""` or `"none"` disables it.
- Out-of-gamut regions are painted with **`--oog-color`** (default: a light/dark-adaptive neutral), with hard edges at the gamut boundaries. The color band itself is left untouched — the overlay is a separate gradient layer.
- **`inGamut`** getter, **`ingamutchange`** event, and **`:state(in-gamut)`** reflecting whether the current color is in gamut (no built-in visual cue — style it via the state).
- **`oogRanges`** getter — the whole-band out-of-gamut ranges as `[start, end]` pairs in 0–1 progress (sorted, merged).
- **`displayGamut`** — the display's detected gamut, i.e. what `gamut="auto"` resolves to.

## `<channel-slider>`

- **`gamut`** property that forwards to its inner `<color-slider>` (`auto` included). `--oog-color` inherits through the shadow boundary, so it needs no forwarding.

## `<color-picker>`

- **`gamut`** property that forwards to every channel slider, so each channel's band marks its out-of-gamut regions. Same values as `<color-slider>`, including `auto`.

## How it works

Gamut membership is a per-pixel property of the *interpolated* color, which CSS can't express, so it's computed in JS. A new `getGamutBoundaries()` util scans each stop-pair interpolation by sampling + bisection — densely for polar interpolations (which can weave in and out of gamut) and at just the endpoints + midpoint for non-polar ones. The per-band ranges are offset into the whole track, merged, rendered as the overlay gradient, and exposed via `oogRanges`.

`background-blend-mode` can't produce transparency and `mask` clips the thumb, which is why a painted overlay (rather than making the band transparent) is the approach.

## Demos

Live examples added to both component READMEs (chroma + hue gamut demos).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #233, opened 2026-06-05
Validate gamut ids

DmitrySharabin:
## Summary
- `getGamutBoundaries()` returns `null` instead of throwing on invalid gamut ids
- `convert()` rejects invalid gamut ids and falls back to the previous value in `color-slider`, `channel-slider`, and `color-picker`

Addresses https://github.com/color-js/elements/pull/231#discussion_r3359282044

## Test plan
- Set `gamut` to an invalid id (e.g. `"bogus"`) → value stays unchanged
- Set `gamut` to valid ids (`"srgb"`, `"p3"`, `"auto"`, `"none"`, `""`) → accepted normally

### DIFF
--- a/src/channel-slider/channel-slider.js
+++ b/src/channel-slider/channel-slider.js
@@ -153,6 +153,12 @@ const Self = class ChannelSlider extends ColorElement {
 		gamut: {
 			type: String,
 			default: "",
+			convert (value) {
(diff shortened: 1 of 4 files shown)


LeaVerou (approved):
I'd swear we had a property specifically for validation but maybe not. LGTM otherwise.

PR #238, opened 2026-06-06
Fix hue-wheel extra points: oversized and not rounded

### LINKED ISSUE #237: hue-wheel: extra points are not rounded and oversized
## Bug

The extra points (slotted `<color-swatch>` and `<color-scale>` swatches) on the `<hue-wheel>` component are broken: they are not circular and are oversized.

Image

## Root cause

In PR #232 (commit 468f20f, "hue-wheel: transparent ring center and plot slotted points"), the original explicit dot styles on `::part(color-swatch)` were removed in favor of delegating to `--swatch-style: dot` via a style container query.

The original code set `width`, `aspect-ratio: 1`, `border-radius: 50%`, `min-inline-size: 0`, `min-block-size: 0` directly on the `::part(color-swatch)` (the host element). The new `@container style(--swatch-style: dot)` block in `color-swatch.css` only targets the inner `#swatch` element — the `:host` element itself never gets compact/circular dot-mode styles.

Additionally, the `<color-scale>` sets `size="large"` on its swatches, which inflates the host to `min(11em, 100%)` × `6em` with `border-radius: 0.2rem` — completely overriding the inner dot sizing.

## Expected

Small, circular dots as in the first version of the component (commit 82f60b4).

## Fix

Add a `@container style(--swatch-style: dot)` block on `:host` in `color-swatch.css` so the host element itself becomes compact and circular in dot mode.

DmitrySharabin:
## Summary

- Override `size="large"` and subgrid styles on `<color-scale>`'s swatches via `::part()` so the dots shrink to fit their content
- Strip inner swatch padding and min-height so dots size to `--point-size` + border

Closes #237

## Test plan

- [x] Verify extra points on the hue-wheel page are small, circular dots
- [x] Hover a dot and confirm the tooltip appears

### DIFF
--- a/src/hue-wheel/hue-wheel-global.css
+++ b/src/hue-wheel/hue-wheel-global.css
@@ -15,6 +15,15 @@ hue-wheel color-scale::part(color-swatch) {
 	top: 50%;
 	left: 50%;
 	margin: 0;
+
+	/* <color-scale> sets size="large" on its swatches and uses subgrid +
+	   containment for even heights. Undo both so the dot shrinks to fit. */
+	display: inline-flex;
+	inline-size: fit-content;
+	min-block-size: 0;
+	contain: none;
+	container-type: normal;
+
 	transition: 150ms scale;
 
 	--_frac: clamp(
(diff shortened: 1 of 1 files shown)


DmitrySharabin:
Fixed in https://github.com/color-js/elements/commit/b6c8070fbd85adddce5703056d42da99231fdca5.

